PR #6, opened 2017-11-14
Added a verification that missing shards are actually missing

jwpgage:
`AtlasMissingShardVerifier` takes the list of missing shards generated by `AtlasShardVerifier`, and checks to see if these shards are missing because they are empty, or because they failed to build. It does this by:

 * Finding the multi polygon of intersection between a shard and country boundary
 * Querying Overpass for all nodes within that bounding box
 * Querying Overpass for all ways within that bounding box
 * Subtracting nodes that are a part of ways that are not ingested into atlas (coastlines, boundaries, etc) from the set of all nodes within the bounding box
 * If any nodes remain within the multipolygon of intersection after filtering, then the shard was not empty and should have been generated
 * The verification returns the list of non empty shards from the missing shards list, along with information about the geometry that proves the shard should have been created.

Also included are classes `OverpassClient`, `OverpassOsmNode`, and `OverpassOsmWay`, which are used query and store information from Overpass respectively.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,6 +1,6 @@
 project.ext.versions = [
     checkstyle: '7.6.1',
-    atlas: '5.0.5-SNAPSHOT',
+    atlas: '5.0.5',
     spark: '1.6.0-cdh5.7.0',
     snappy: '1.1.1.6',
 ]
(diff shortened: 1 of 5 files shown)


MikeGost:
@jwpgage has more changes coming to this PR, so holding off on a merge for now.

MikeGost:
This needs the latest Atlas release artifact to compile properly.

PR #7, opened 2017-11-14
File System Helper and HDFS walker update

jklamer:
These changes allow the file system helper to create a HDFS walker using a file system configuration as well as stream the paths in the file system to look for files that have information stored in their relative path. This PR also updates relevant testing. Credit to [cstaylor](https://github.com/cstaylor) for all of the work regarding file system configuration and updating relevant testing and usages.


src/test/java/org/openstreetmap/atlas/generator/tools/streaming/resource/HDFSWalkerTest.java
```diff
@@ -1,93 +0,0 @@
...
-import org.junit.Test;
-import org.openstreetmap.atlas.streaming.resource.File;
-import org.openstreetmap.atlas.utilities.tuples.Tuple;
-
-/**
- * Tests for {@link HDFSWalker}.
- *
- * @author mkalender
- */
-public class HDFSWalkerTest
```
MikeGost: Remind me why we're getting rid of this class?

PR #10, opened 2018-01-13
HDFSWalker should take in max depth for traversal

snb3300:
Current implementation of HDFSWalker traverses all the files and directories in the root using breadth first traversal. Limiting this traversal can have an advantage of filtering out unnecessary directory explorations beyond certain depth and gain some performance. This PR addresses the issue and the consumer of this class can limit the file hierarchy exploration beyond certain depth.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/tools/streaming/resource/HDFSWalker.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/tools/streaming/resource/HDFSWalker.java
@@ -22,8 +22,8 @@ import com.google.common.collect.AbstractIterator;
 /**
  * Inspired by Files.walk
  * (https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#walk-java.nio.file.Path-int-
- * java.nio.file.FileVisitOption...-), either retrieve a Stream of HDFS file
- * status objects or convert them to HDFSFile objects.
+ * java.nio.file.FileVisitOption...-), either retrieve a Stream of HDFS file status objects or
+ * convert them to HDFSFile objects.
  *
  * @author cstaylor
  * @author jklamer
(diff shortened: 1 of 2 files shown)


src/test/java/org/openstreetmap/atlas/generator/tools/streaming/resource/HDFSWalkerTest.java
```diff
@@ -14,27 +14,60 @@
...
+        final File rootDirectory = File.temporaryFolder();
+        final File pathOne = rootDirectory.child("test-a");
+        final File pathTwo = rootDirectory.child("test-b");
+        pathOne.mkdirs();
+        pathTwo.mkdirs();
+        pathOne.child("test-a.tmp").writeAndClose("test file");
+        pathTwo.child("test-b.tmp").writeAndClose("test file");
+
+        // This should return only 2 child directories of the rootDirectory
+        Tuple<List<FileStatus>, List<String>> results = HDFSWalkerTest.test(rootDirectory, 1);
```
MikeGost: Nice tests! Any value in adding one that tests depth of 0, or only root?

src/main/java/org/openstreetmap/atlas/generator/tools/streaming/resource/HDFSWalker.java
```diff
@@ -69,24 +118,37 @@ protected FileStatus computeNext()
...
-                    throw new CoreException("Can't locate children of {}", returnValue, oops);
+                    try
+                    {
+                        Stream.of(this.fileSystem.listStatus(currentFileStatus.getPath()))
+                                .forEach(status -> this.currentPaths
+                                        .add(new FileStatusAtDepth(status, childDepth)));
+                    }
+                    catch (final IOException oops)
+                    {
+                        throw new CoreException("Can't locate children of {}", currentFileStatus,
```
MikeGost: Do we want to log the depth too? Something like, "... children of {} at depth {}"

src/main/java/org/openstreetmap/atlas/generator/tools/streaming/resource/HDFSWalker.java
```diff
@@ -27,33 +27,82 @@
...
+            this(root, configuration, HDFSWalker.WALK_ALL);
+        }
+
+        private HDFSIterator(final Path root, final Configuration configuration, final int maxDepth)
         {
             if (root == null)
             {
                 throw new CoreException("Error when creating an HDFSIterator: root can't be null");
             }
+            this.maxDepth = maxDepth;
```
MikeGost: Was thinking edge cases, I guess parameter validation doesn't apply here. If we're passed in -100, it would be the same as having no depth.

MikeGost:
Didn't realize this was already merged. Added some feedback here and there, feel free to address in a new PR or refute :)

src/main/java/org/openstreetmap/atlas/generator/tools/streaming/resource/HDFSWalker.java
```diff
@@ -27,33 +27,82 @@
...
+     * hierarchy
+     *
+     * @author sbhalekar
+     */
+    private static final class FileStatusAtDepth
+    {
+        private final int fileDepth;
+        private final FileStatus fileStatus;
+
+        private FileStatusAtDepth(final FileStatus fileStatus, final int fileLevel)
```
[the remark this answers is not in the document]
snb3300: makes sense

src/main/java/org/openstreetmap/atlas/generator/tools/streaming/resource/HDFSWalker.java
snb3300: Yes, cases like negative depth or depth values more than the actual depth would be same as having no depth

src/main/java/org/openstreetmap/atlas/generator/tools/streaming/resource/HDFSWalker.java
snb3300: I can add it If you think that's helpful

PR #11, opened 2018-02-06
Add support for binaryFile in SparkInput

matthieun:
This PR upgrades `SparkInput` to support binary files the same way it supports sequence files.

The only requirement on the underlying file system is to efficiently return the length of a file, as the binary reader cannot infer the length of a file in advance.

See the new test for an example use case.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/tools/spark/input/SparkInput.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/tools/spark/input/SparkInput.java
@@ -20,12 +20,17 @@ import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.SequenceFile;
 import org.apache.hadoop.io.SequenceFile.Reader.Option;
 import org.apache.hadoop.io.Writable;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/generator/tools/spark/input/SparkInput.java
```diff
@@ -90,6 +95,49 @@
...
+                    }
+                    final CombineFileSplit split = new CombineFileSplit(new Path[] { elasticPath },
+                            new long[] { fileLength });
+                    final PortableDataStream result = new PortableDataStream(split,
+                            taskAttemptContext, 0);
+                    return Iterables.from(new Tuple2<>(elasticPath.toString(), result));
+                },
+                // In case of a non elastic file system, like HDFS, use the standard method.
+                otherPath -> context.binaryFiles(otherPath));
+    }
```
sidr: Nicely done. This is a useful addition !!

jwpgage:
Looks good!

PR #12, opened 2018-02-14
Added check for too much data in AtlasMissingShardVerifier

jwpgage:
Adds a check to ensure that the overpass response is not empty due to requesting too much data. In this case, there is definitely something missing, and therefore the check fails. However, we won't know exactly which shards failed. There are deeper solutions to this that I might explore at a future date, but this is a solution that removes the false positive that is currently returned.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/sharding/OverpassClient.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/sharding/OverpassClient.java
@@ -24,7 +24,7 @@ import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
 
 /**
- * @author james-gage
+ * @author jwpgage
  */
 public class OverpassClient
 {
--- a/src/main/java/org/openstreetmap/atlas/generator/sharding/OverpassClient.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/sharding/OverpassClient.java
@@ -1,7 +1,6 @@
 package org.openstreetmap.atlas.generator.sharding;
 
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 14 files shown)


src/main/java/org/openstreetmap/atlas/generator/sharding/OverpassClient.java
```diff
@@ -164,8 +180,9 @@ public CloseableHttpResponse getResponse(final String specificQuery)
         try (CloseableHttpResponse response = this.getResponse(query))
         {
             logger.info("Parsing node query response...");
-            final Document doc = builder.parse(response.getEntity().getContent());
-            final NodeList nodelist = doc.getElementsByTagName("node");
+            final Document document = builder.parse(response.getEntity().getContent());
+            checkForTooMuchData(document);
+            final NodeList nodelist = document.getElementsByTagName("node");
```
MikeGost: If we see too much data returned, does it even make sense to proceed with the code below? What are your thoughts on bailing out early - maybe throwing Exception or returning empty list.

src/main/java/org/openstreetmap/atlas/generator/sharding/OverpassClient.java
```diff
@@ -59,6 +60,16 @@ public static String buildQuery(final String type, final String key, final Strin
                 + END_QUERY;
     }
 
+    private static void checkForTooMuchData(final Document document)
+    {
+        if (document.getDocumentElement().getElementsByTagName("remark").item(0) != null
+                && document.getDocumentElement().getElementsByTagName("remark").item(0)
+                        .getTextContent().contains("runtime error: Query ran out of memory"))
```
matthieun: Is it possible to catch any type of error here? That would make the verification tool more robust, right?

jwpgage: Since it's not a response code that we read but instead the contents of the returned xml document, I'd have to see multiple errors and look for common keywords to catch for. I suppose we could fail if we just see the word error in the remark tag, but for that case we would have to write a more general exception message.

PR #13, opened 2018-02-22
Moving SparkFileHelper and dependencies from atlas-checks to atlas-generator

AlexHsieh22:
SparkFileHelper and dependencies is currently found in [atlas-checks](https://github.com/osmlab/atlas-checks/tree/dev/src/main/java/org/openstreetmap/atlas/checks/persistence). It is a set of utilities that simplifies Hadoop FileSystem interactions with atlas files.

Based on that, the right home for these files is within atlas-generator. After this has been merged, I'll propose a corresponding PR in atlas-checks to update the dependency and remove the files.


MikeGost:
@AlexHsieh22 - can we expect a corresponding PR in atlas-checks to remove the migrated code?

AlexHsieh22:
Yes! I'll get one up soon!

PR #14, opened 2018-02-23
Distribute atlas generator country to shard mapping

matthieun:
This change parallelizes the `AtlasGenerator`'s initial step of figuring out the shards that are covered by a country boundary. It is now a two-step process, one with finding all the shards that intersect the country boundary's rectangular bounds, then another parallel process where each shard is filtered out if it does not actually cover the bounds.

This is a temporary performance update until the `CountryBoundaryMap`'s `gridIndex` can be used to speed up the process.


MikeGost:
Also, do you think a quick TO-DO might be useful to say... "leverage grid index" once the changes are ready?

src/test/java/org/openstreetmap/atlas/generator/AtlasGeneratorTest.java
```diff
@@ -0,0 +1,57 @@
...
+                new InputStreamResource(
+                        () -> AtlasGeneratorTest.class.getResourceAsStream("JAM.txt")));
+
+        final List<CountryBoundary> countryBoundaryHTI = countryBoundaryMapHTI
+                .countryBoundary("HTI");
+        final List<CountryBoundary> countryBoundaryJAM = countryBoundaryMapJAM
+                .countryBoundary("JAM");
+
+        final Sharding sharding = new DynamicTileSharding(new InputStreamResource(
+                () -> AtlasGenerator.class.getResourceAsStream("tree-6-14-100000.txt")));
```
MikeGost: The sharding tree already exists in src/test/resources/org/openstreetmap/atlas/geography/boundary/tree-6-14-100000.txt.gz - let's reuse that!

matthieun: So that is in another project (atlas), and unfortunately the test resources are not packaged in the library jar I don't think...

MikeGost: ohhh good point! Forgot that this is atlas-generator, not atlas - sorry!

MikeGost:
Merging for a release! Thanks @matthieun

PR #18, opened 2018-03-08
Move country-shard generation to Spark master

mertemin:
This PR updates `AtlasGenerator` to generate `Shard` tasks in Spark master rather than doing it through Spark executors. This process doesn't need to be distributed over executors, because all the information needed for this operation is already available in the master and also the object used for this operation `CountryBoundaryMap` instance might not be not small. Passing `CountryBoundaryMap` instance back and forth adds extra cost and slows down the generation process.

- To make things faster, initial parallelization `JavaRDD` and `JavaPairRDD`s are removed and replaced with `generateTasks` method call. This method goes and generates `AtlasGenerationTask`s to be used in the generation process.
- `AtlasGenerationTaskProcessor` is created to generate tasks in parallel.
- Methods `roughShards` and `filterShards` are removed. Rather, `Sharding`'s `shards` method is being used. See https://github.com/osmlab/atlas/blob/0800d37ccff186d70723e6727934c91e1ff1ecc5/src/main/java/org/openstreetmap/atlas/geography/sharding/Sharding.java#L88 for implementation details.
- `PbfLoader`'s constructor is updated to accept `List` of `Shard`s for simplicity and reuse from `AtlasGenerator` side.
- `AtlasGeneratorTest` is updated with new tests.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
@@ -3,7 +3,6 @@ package org.openstreetmap.atlas.generator;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -43,6 +42,7 @@ import org.openstreetmap.atlas.tags.filters.ConfiguredTaggableFilter;
 import org.openstreetmap.atlas.utilities.collections.StringList;
 import org.openstreetmap.atlas.utilities.configuration.StandardConfiguration;
 import org.openstreetmap.atlas.utilities.conversion.StringConverter;
+import org.openstreetmap.atlas.utilities.maps.MultiMapWithSet;
 import org.openstreetmap.atlas.utilities.runtime.CommandMap;
 import org.openstreetmap.atlas.utilities.runtime.system.memory.Memory;
 import org.openstreetmap.atlas.utilities.threads.Pool;
@@ -149,7 +149,7 @@ public class AtlasGenerator extends SparkJob
     {
         // Extract country boundaries and queue them
         final BlockingQueue<CountryBoundary> queue = new LinkedBlockingQueue<>();
-        final Map<String, Set<Shard>> countryToShardMap = new HashMap<>();
+        final MultiMapWithSet<String, Shard> countryToShardMap = new MultiMapWithSet<>();
         countries.stream().forEach(country ->
         {
             // Initialize country-shard map
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
@@ -20,6 +20,7 @@ import org.openstreetmap.atlas.generator.persistence.MultipleAtlasStatisticsOutp
 import org.openstreetmap.atlas.generator.persistence.delta.RemovedMultipleAtlasDeltaOutputFormat;
 import org.openstreetmap.atlas.generator.sharding.AtlasSharding;
 import org.openstreetmap.atlas.generator.tools.spark.SparkJob;
+import org.openstreetmap.atlas.geography.Rectangle;
 import org.openstreetmap.atlas.geography.atlas.Atlas;
 import org.openstreetmap.atlas.geography.atlas.delta.AtlasDelta;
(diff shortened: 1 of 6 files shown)


mertemin:
Marking this as DNM as I might have much faster approach leveraging `CountryBoundaryMap`'s grid index.

matthieun:
Closing until it is ready. We can still comment!

matthieun:
Very nice test cleanup, thanks @mertemin

mertemin:
Let's reopen this @matthieun I pushed a new commit to the branch that should make things faster.

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorTaskProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+            final String country = item.getCountryName();
+
+            // Skip a shard if
+            // - there is no corresponding grid for given country
+            // - or given country boundary doesn't overlap with shard bounds
+            final Rectangle shardBounds = shard.bounds();
+            if (this.boundaryMap.countryCodesOverlappingWith(shardBounds).contains(country)
+                    && item.getBoundary().overlaps(shardBounds))
+            {
+                synchronized (this.countryToShardMap)
```
MikeGost: Random thought here - any value in adding a ConcurrentMultiMap implementation (backed by a ConcurrentHashMap) instead of synchornizing here. I don't think it would be much gain, just tossing an idea out.

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorTaskProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+import org.openstreetmap.atlas.geography.boundary.CountryBoundary;
+import org.openstreetmap.atlas.geography.boundary.CountryBoundaryMap;
+import org.openstreetmap.atlas.geography.sharding.Shard;
+import org.openstreetmap.atlas.geography.sharding.Sharding;
+import org.openstreetmap.atlas.utilities.maps.MultiMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Worker that acts like both producer and consumer. Takes an {@link CountryBoundary} from work
```
MikeGost: Nit: an --> a

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorTaskProcessor.java
mertemin: Let me try using `ConcurrentMultiMap` and see how it affects the performance and results.

mertemin: Tried `ConcurrentMultiMap` from `akka.util` package. Using it requires certain plumbing to make it work with it's interface, which is somewhat different (see [this](https://doc.akka.io/api/akka/current/akka/util/ConcurrentMultiMap.html)). However, while doing that I realized `MultiMap` can have duplicate shards (in extreme cases of country boundaries with duplicate parts). So I changed it to `Map<String, Set<Shard>>` and starting locking only country specific shard set, not the whole data structure. That improved performance a little more.

mertemin: I'll address duplicate boundary issue in `CountryBoundaryMap` as well.

MikeGost: Nice find!

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -131,65 +133,71 @@ public static void main(final String[] args)
...
      */
-    protected static boolean filterShards(final Shard shard, final List<CountryBoundary> boundaries)
+    protected static List<AtlasGenerationTask> generateTasks(final StringList countries,
+            final CountryBoundaryMap boundaryMap, final Sharding sharding)
     {
-        boolean result = false;
-        for (final CountryBoundary boundary : boundaries)
+        // Extract country boundaries and queue them
+        final BlockingQueue<CountryBoundary> queue = new LinkedBlockingQueue<>();
+        final Map<String, Set<Shard>> countryToShardMap = new HashMap<>();
```
MikeGost: Would a`MultiMapWithSet` work here instead?

mertemin: Yeah.

PR #22, opened 2018-03-21
Create a new spark task per shard

matthieun:
This PR is to ensure each shard is processed in a separate task.


matthieun:
@mertemin I don't think it should always be the case. Here I just wanted to add it so there is one shard per task which for me makes it easier to debug and identify issues, especially related to memory issues I find. If there is a case against having one shard per task, I am ready to hear it out!

mertemin:
@matthieun Sure, that sounds fair.

PR #25, opened 2018-04-06
Do Not Merge: Raw Atlas Flow Integration

MikeGost:
This build is expected to fail, since it depends on [this](https://github.com/osmlab/atlas/pull/106) Atlas PR.

The main change here is to integrate the Raw Atlas generation flow and make it backward compatible. This is done by to adding a "useRawAtlas" `Flag`, which when present will generate an Atlas using the Raw Atlas flow. I've tested this (both with and without) the flag successfully.

Next steps include:
1. Adding Raw Atlas Fetcher for the sectioning step
2. Adding Metrics/Statistics for the final Atlas
3. General cleanup 
4. Parity tests

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
@@ -27,6 +27,8 @@ import org.openstreetmap.atlas.generator.tools.spark.SparkJob;
 import org.openstreetmap.atlas.geography.atlas.Atlas;
 import org.openstreetmap.atlas.geography.atlas.delta.AtlasDelta;
 import org.openstreetmap.atlas.geography.atlas.pbf.AtlasLoadingOption;
+import org.openstreetmap.atlas.geography.atlas.raw.sectioning.WaySectionProcessor;
+import org.openstreetmap.atlas.geography.atlas.raw.slicing.RawAtlasCountrySlicer;
 import org.openstreetmap.atlas.geography.atlas.statistics.AtlasStatistics;
 import org.openstreetmap.atlas.geography.atlas.statistics.Counter;
 import org.openstreetmap.atlas.geography.boundary.CountryBoundary;
@@ -35,6 +37,7 @@ import org.openstreetmap.atlas.geography.boundary.CountryBoundaryMapArchiver;
 import org.openstreetmap.atlas.geography.sharding.CountryShard;
 import org.openstreetmap.atlas.geography.sharding.Shard;
 import org.openstreetmap.atlas.geography.sharding.Sharding;
+import org.openstreetmap.atlas.locale.IsoCountry;
 import org.openstreetmap.atlas.streaming.resource.Resource;
 import org.openstreetmap.atlas.tags.filters.ConfiguredTaggableFilter;
 import org.openstreetmap.atlas.utilities.collections.StringList;
@@ -48,16 +51,17 @@ import org.openstreetmap.atlas.utilities.time.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.Maps;
+
 import scala.Tuple2;
 
 /**
- * Generate {@link Atlas} Shards for a specific version and a specific set of Countries
+ * Generate {@link Atlas} Shards for a specific version and a specific set of countries
  *
  * @author matthieun
  */
 public class AtlasGenerator extends SparkJob
 {
-
     /**
      * @author matthieun
      */
@@ -77,6 +81,8 @@ public class AtlasGenerator extends SparkJob
 
     private static final Logger logger = LoggerFactory.getLogger(AtlasGenerator.class);
     public static final String ATLAS_FOLDER = "atlas";
+    public static final String RAW_ATLAS_FOLDER = "rawAtlas";
+    public static final String SLICED_RAW_ATLAS_FOLDER = "slicedRawAtlas";
     public static final String SHARD_STATISTICS_FOLDER = "shardStats";
     public static final String COUNTRY_STATISTICS_FOLDER = "countryStats";
     public static final String SHARD_DELTAS_FOLDER = "deltas";
@@ -129,6 +135,9 @@ public class AtlasGenerator extends SparkJob
             "osmPbfRelationConfiguration",
             "The path to the configuration file that defines which PBF Relations becomes an Atlas Entity",
             StringConverter.IDENTITY, Optionality.OPTIONAL);
+    // TODO - once this is fully baked and tested, remove flag and old flow
+    private static final Flag USE_RAW_ATLAS = new Flag("useRawAtlas",
+            "Allow PBF to Atlas process to use Raw Atlas flow");
 
     public static void main(final String[] args)
     {
@@ -167,7 +176,7 @@ public class AtlasGenerator extends SparkJob
             }
 
             // Queue boundary for processing
-            countryBoundaries.forEach(countryBoundary -> queue.add(countryBoundary));
+            countryBoundaries.forEach(queue::add);
         });
 
         // Use all available processors except one (used by main thread)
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -284,243 +360,428 @@ public void start(final CommandMap command)
...
+            // from the command map and pass those around to create the AtlasLoadingOption
+            final Map<String, String> atlasLoadingOptions = extractAtlasLoadingProperties(command);
+
+            // Generate the raw Atlas and save it
+            final JavaPairRDD<String, Atlas> countryRawAtlasShardsRDD = generateAndSaveRawAtlas(
+                    tasks, boundaries, sparkContext, atlasLoadingOptions, pbfContext, output,
+                    codeVersion, dataVersion);
+
+            // Slice the raw Atlas and save result
+            final JavaPairRDD<String, Atlas> countrySlicedRawAtlasShardsRDD = sliceAndSaveRawAtlases(
```
matthieun: To make it easier to read the Spark flow, it would be nicer to see the "mapToPair" "filter", "reduce" calls here directly.

What I would do instead here is make the `generateAndSaveRawAtlas` method return a `Function` that does the transformation for you between the `Tuples`. That way all you have to do here is something that will look like this:
```
final JavaPairRDD<String, Atlas> countrySlicedRawAtlasShardsRDD = 
    countryRawAtlasShardsRDD.mapToPair(sliceAndSaveRawAtlases(.....));
```

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -284,243 +360,428 @@ public void start(final CommandMap command)
...
                     // Report on memory usage
-                    logger.info("Printing memory after loading Atlas {}", name);
+                    logger.info("Printing memory after loading raw Atlas {}", name);
                     Memory.printCurrentMemory();
+
                     // Output the Name/Atlas couple
                     final Tuple2<String, Atlas> result = new Tuple2<>(name, atlas);
                     return result;
-                });
+                }).filter(tuple -> tuple._2() != null);
```
matthieun: You could also bubble up that filter call at the top level, for the same reasons.

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -284,243 +360,428 @@ public void start(final CommandMap command)
...
 
-        // Filter out null Atlas.
-        final JavaPairRDD<String, Atlas> countryNonNullAtlasShardsRDD = countryAtlasShardsRDD
-                .filter(tuple -> tuple._2() != null);
+        // Persist the RDD to avoid re-computing and save each raw atlas
+        countryRawAtlasShardsRDD.cache().saveAsHadoopFile(
```
matthieun: Same, bubble this up at the top level. `cache()` calls are integral to understanding the Spark flow, and I would love to see them in the top method..

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -284,243 +360,428 @@ public void start(final CommandMap command)
...
                 .mapToPair(tuple ->
                 {
-                    final Counter counter = new Counter().withSharding(sharding);
-                    counter.setCountsDefinition(Counter.POI_COUNTS_DEFINITION.getDefault());
-                    final AtlasStatistics statistics;
+                    final Atlas atlas;
                     try
                     {
-                        statistics = counter.processAtlas(tuple._2());
+                        final AtlasLoadingOption atlasLoadingOption = buildAtlasLoadingOption(
```
matthieun: I am not sure I remember how the legacy code passes the `AtlasLoadingOption` around, but I believe it did not have to do this. We can review this at some point, there might be another way :)

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
MikeGost: Fair point, I'll update this!

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
MikeGost: I don't think it passes it around, each slave builds it's own AtlasLoadingOption! Building it once and passing it around didn't work, since it will try to serialize the entire class and its contents - some of which aren't Serializable.

MikeGost: I'll try to think of something a little more elegant, if you have ideas, please let me know!

MikeGost:
This now relies on this Atlas [PR](https://github.com/osmlab/atlas/pull/108).

MikeGost:
This PR is feature complete. It needs a new release of the Atlas project before it can be merged. I've tested and visualized Atlas Generation with and without the new flag - it works as expected. I've also ran the integration test with the new flag, it passed!

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -217,6 +231,87 @@ protected static StandardConfiguration getStandardConfigurationFrom(final String
...
+                (String) command.get(PBF_WAY_CONFIGURATION));
+        propertyMap.put(PBF_RELATION_CONFIGURATION.getName(),
+                (String) command.get(PBF_RELATION_CONFIGURATION));
+        propertyMap.put(CODE_VERSION.getName(), (String) command.get(CODE_VERSION));
+        propertyMap.put(DATA_VERSION.getName(), (String) command.get(DATA_VERSION));
+
+        return propertyMap;
+    }
+
+    private static Set<Shard> getAllShardsForCountry(final CountryBoundaryMap boundaries,
```
matthieun: Isn't that going to be slow? It would probably benefit from using the grid index, like there:
https://github.com/osmlab/atlas-generator/blob/4.0.4/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java#L138-L204

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -295,222 +392,266 @@ public void start(final CommandMap command)
...
+            // Persist the RDD and save the intermediary state
+            countryRawAtlasShardsRDD.cache();
+            countryRawAtlasShardsRDD.saveAsHadoopFile(
+                    getAlternateSubFolderOutput(output, RAW_ATLAS_FOLDER), Text.class, Atlas.class,
+                    MultipleAtlasOutputFormat.class, new JobConf(configuration()));
+            logger.info("\n\n********** SAVED THE RAW ATLAS **********\n");
+
+            // Slice the raw Atlas and filter any null atlases
+            final JavaPairRDD<String, Atlas> countrySlicedRawAtlasShardsRDD = countryRawAtlasShardsRDD
+                    .mapToPair(this.sliceRawAtlas(boundaries)).filter(tuple -> tuple._2() != null);
```
matthieun: Here passing the boundaries means the driver will have to serialize it and pass it around to all the slaves. I guess the old flow does the same thing, not sure if there is a better way to do that... Would it be possible to only pass the boundary of the country that is being sliced? Not sure.

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -295,222 +392,266 @@ public void start(final CommandMap command)
...
+
+            // Persist the RDD and save the intermediary state
+            final String slicedRawAtlasPath = getAlternateSubFolderOutput(output,
+                    SLICED_RAW_ATLAS_FOLDER);
+            countrySlicedRawAtlasShardsRDD.cache();
+            countrySlicedRawAtlasShardsRDD.saveAsHadoopFile(slicedRawAtlasPath, Text.class,
+                    Atlas.class, MultipleAtlasOutputFormat.class, new JobConf(configuration()));
+            logger.info("\n\n********** SAVED THE SLICED RAW ATLAS **********\n");
```
matthieun: Here I believe you can now release the `countryRawAtlasShardsRDD`, because the next step, `countrySlicedRawAtlasShardsRDD` has already been cached.

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -295,222 +392,266 @@ public void start(final CommandMap command)
...
+            final JavaPairRDD<String, Atlas> countryAtlasShardsRDD = countrySlicedRawAtlasShardsRDD
+                    .mapToPair(this.sectionRawAtlas(boundaries, sharding, sparkContext,
+                            atlasLoadingOptions, slicedRawAtlasPath));
+
+            // Persist the RDD and save the final atlas
+            countryAtlasShardsRDD.cache();
+            countryAtlasShardsRDD.saveAsHadoopFile(
+                    getAlternateSubFolderOutput(output, ATLAS_FOLDER), Text.class, Atlas.class,
+                    MultipleAtlasOutputFormat.class, new JobConf(configuration()));
+            logger.info("\n\n********** SAVED THE FINAL ATLAS **********\n");
```
matthieun: Same here, `countrySlicedRawAtlasShardsRDD` should be releasable here.

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -295,222 +392,266 @@ public void start(final CommandMap command)
...
+            final JavaPairRDD<String, AtlasStatistics> statisticsRDD = countryAtlasShardsRDD
+                    .mapToPair(generateAtlasStatistics(sharding));
+
+            // Persist the RDD and save
+            statisticsRDD.cache();
+            statisticsRDD.saveAsHadoopFile(
+                    getAlternateSubFolderOutput(output, SHARD_STATISTICS_FOLDER), Text.class,
+                    AtlasStatistics.class, MultipleAtlasStatisticsOutputFormat.class,
+                    new JobConf(configuration()));
+            logger.info("\n\n********** SAVED THE SHARD STATISTICS **********\n");
```
matthieun: And again, same here. Although at that point in the job, releasing RDD memory might not make much sense anymore.

matthieun: Actually not here, as it is still used by the Deltas below!

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -534,6 +675,242 @@ protected SwitchList switches()
...
+
+    /**
+     * @param sparkContext
+     *            Spark context (or configuration) as a key-value map
+     * @param previousOutputForDelta
+     *            Previous Atlas generation delta output location
+     * @return A Spark {@link PairFlatMapFunction} that takes a tuple of a country shard name and
+     *         atlas file and returns all the {@link AtlasDelta} for the country
+     */
+    private PairFlatMapFunction<Tuple2<String, Atlas>, String, AtlasDelta> computeAtlasDelta(
```
matthieun: Make all those functions static so the whole SparkJob does not have to be serialized and passed around for the slaves to be able to call those methods!

matthieun: And even better: push them to the helper!

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -0,0 +1,175 @@
...
+import org.openstreetmap.atlas.streaming.resource.File;
+import org.openstreetmap.atlas.streaming.resource.FileSuffix;
+import org.openstreetmap.atlas.streaming.resource.Resource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Utility class for {@link AtlasGenerator}.
+ *
+ * @author matthieun
```
matthieun: Did I write that class??

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
MikeGost: Yes - that's a good point, let me update to use this!

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
MikeGost: That's a good point, let me think of a better way to handle this or at least create an issue!

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
MikeGost: Good call, let me do this!

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
MikeGost: You inspired it :)

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
MikeGost: Knew I forgot something, good catch!!

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
MikeGost: Created: https://github.com/osmlab/atlas-generator/issues/28 - let's come back to this. I think it's important, but something we can live with in the meantime.

PR #21, opened 2018-03-19
Update PBF input configuration and allow separate sharding

matthieun:
- Allow for a separate sharding for the input PBF files
- Allow for a configurable folder structure (called scheme) for the input PBF files
- Cleanup `PbfLocatorTest` and remove old PBF test resource
- Make the output format of shard-based Atlas and Statistics files more granular
  - `COUNTRY/Z/X/Y/COUNTRY_Z-X-Y.atlas`

PR #26, opened 2018-04-06
Allow configurable Atlas output folder structure

matthieun:
This PR is somehow related to https://github.com/osmlab/atlas-generator/pull/21

This adds a new switch to `AtlasGenerator` that allows the user to configure the output folder structure for Atlas files. For example:
```
-atlasScheme=zz/
```
will have the output atlas files be separated by zoom folders:
```
output/atlas/COUNTRY/Z/COUNTRY_Z-X-Y.atlas
```
By default, all the output files are in the same country folder.

Also:

- Fixed an issue in `ResourceFileSystem` where Spark renaming of deep folder result files was not working properly
- Added an integration test that builds atlas shards for a very small cutout of Dominica.


src/main/java/org/openstreetmap/atlas/generator/tools/streaming/ResourceFileSystem.java
```diff
@@ -184,7 +196,21 @@ public boolean rename(final Path source, final Path destination) throws IOExcept
             // Hadoop end of job rename case
             destinationName = destinationName.endsWith("/") ? destinationName
                     : destinationName + "/";
-            final String appendName = sourceName.substring(sourceName.lastIndexOf("/") + 1);
+            // Sample resource:
+            // resource://test/atlas/_temporary/0/_temporary/attempt_201804061442_0003_m_000002_14/DMA/9/DMA_9-168-234.atlas
+            final Pattern pattern = Pattern.compile("\\/attempt_[^\\/]*\\/");
```
jwpgage: Nice regex :)

src/main/java/org/openstreetmap/atlas/generator/tools/streaming/ResourceFileSystem.java
```diff
@@ -184,7 +196,21 @@ public boolean rename(final Path source, final Path destination) throws IOExcept
...
+            final Matcher matcher = pattern.matcher(sourceName);
+            final String appendName;
+            if (matcher.find())
+            {
+                final String match = matcher.group(0);
+                final StringList split = StringList.split(sourceName, match);
+                appendName = split.get(1);
```
jwpgage: Is there any chance the folder could be empty, and .get(1) could return IndexOutOfBounds?

src/integrationTest/java/org/openstreetmap/atlas/generator/AtlasGeneratorIntegrationTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.openstreetmap.atlas.utilities.collections.StringList;
+
+/**
+ * @author matthieun
+ */
+public class AtlasGeneratorIntegrationTest
+{
+    public static final String BOUNDARY = "resource://test/boundaries/DMA.txt";
+    public static final String TREE_13 = "resource://test/sharding/tree-6-13-100000.txt";
+    public static final String TREE_14 = "resource://test/sharding/tree-6-14-100000.txt";
```
MikeGost: Is it possible to re-use `tree-6-14-100000.txt` from `src/test/resources/org/openstreetmap/atlas/generator` to avoid having two large identical files?

matthieun: Let me try that.

src/main/java/org/openstreetmap/atlas/generator/tools/streaming/ResourceFileSystem.java
matthieun: This `ResourceFileSystem` is used for tests mostly. Here it is doing a big hack to conform to the Spark file naming, and yes if they change it it will break. Not sure there is a better way right now..

PR #29, opened 2018-04-14
Use switch instead of flag for consistency

matthieun:
All other parameters are Switches. This is to align `useRawAtlas` with the others.


MikeGost:
Why is this needed? A Flag is simpler version of a Boolean Switch.

PR #35, opened 2018-05-08
Checking for Remote Atlas File Existence When Fetching

MikeGost:
Two changes here:

1. We try to reach out to a remote source to get the corresponding `Atlas` file for a given `Shard`. However, it's perfectly valid that there is no `Atlas` file for that `Shard`. If this is the case, we were throwing and bubbling the Exception up the stack. This change looks for a `FileNotFoundException` and returns and `Optional.empty()` for the requested `Shard`. All other behavior is the same.
2. Cleaning up `system.println` statements and updating them to use logger.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
@@ -122,7 +122,26 @@ public final class AtlasGeneratorHelper implements Serializable
 
                     // FileSystemHelper.resource sets the Decompressor on the Resource for us, so
                     // this call will gunzip the file
-                    fileFromNetwork.copyTo(temporaryLocalFile);
+                    try
+                    {
+                        fileFromNetwork.copyTo(temporaryLocalFile);
+                    }
+                    catch (final Exception e)
+                    {
+                        final Optional<FileNotFoundException> fileNotFound = ExceptionSearch
+                                .find(FileNotFoundException.class).within(e);
+                        if (fileNotFound.isPresent())
+                        {
+                            // It's possible there is no Atlas file for a given shard
+                            logger.info("No Atlas file found at {}", path);
+                            return Optional.empty();
+                        }
+                        else
+                        {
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -115,12 +117,31 @@
...
                     // FileSystemHelper.resource sets the Decompressor on the Resource for us, so
                     // this call will gunzip the file
-                    fileFromNetwork.copyTo(temporaryLocalFile);
+                    try
+                    {
+                        fileFromNetwork.copyTo(temporaryLocalFile);
+                    }
+                    catch (final Exception e)
+                    {
+                        final Optional<FileNotFoundException> fileNotFound = ExceptionSearch
```
lucaspcram: Quick question Mike. Is there an advantage to using this `Optional` approach versus just doing something like:

```Java
try
{
    // code
}
catch (FileNotFoundException exception)
{
    // FileNotFoundException handling here
}
catch(Exception exception)
{
    // anything else falls down to here
}
```

I am curious if the `Optional` approach is providing some additional functionality that the above approach misses. Thanks!

MikeGost: Great callout @lucaspcram! I wish we could do that, that would be a lot more straightforward. The reason for using `ExceptionSearch` here is that the `FileNotFoundException` happens at the very bottom of the stack - Hadoop FileSystem default behavior. That's then wrapped by a number of CoreExceptions in our code - the full stack trace had 3 or 4 layers. So we would have to walk the entire stack and find the root cause of the Exception and determine if it was FileNotFoundException and handle it accordingly. This is what `ExceptionSearch` is doing here, courtesy of @cstaylor.

lucaspcram: Ah wow ok that's hairy. I checked out the `ExceptionSearch` class. Very clever @cstaylor

PR #36, opened 2018-05-11
DNM: Synchronize by country instead of raw atlas stage

MikeGost:
Previous implementation would synchronize across stages - once all raw atlases were built, then all slicing could start. Once all slicing would be completed, all sectioning could start. However, this didn't make sense - why does a small country that has relatively few shards have to wait for a large country with orders of magnitude more shards than it to complete? This PR updates the the synchronization of Spark stages (raw atlas creation, slicing, sectioning, statistics) to happen at the country level. As a result this allows more efficient executor utilization. Each executor can start be processing something instead of waiting for the last few shards of some stage to complete. 

DNM: This has encountered an issue when tested extensively at scale.


MikeGost:
Going to close this for now, as it's causing issues at scale. This needs more work.

PR #42, opened 2018-07-14
Build updates: Sonar, Jacoco, Gradle 4.8.1

matthieun:
Update to the build scripts and sonat/jacoco integration.


build.gradle
```diff
@@ -1,24 +1,21 @@
...
+plugins {
+	id 'java'
+	id 'maven'
+	id 'idea'
+	id 'signing'
+	id 'checkstyle'
+	id 'jacoco'
+	id "com.diffplug.gradle.spotless" version "3.4.0"
+	id 'org.sonarqube' version '2.6'
+	id 'com.google.protobuf' version '0.8.1'
```
MikeGost: Why are we introducing protobuf plugin here?

matthieun: Good catch, this is a copy paste error. Will remove it in a later PR as this is harmless.

PR #43, opened 2018-07-18
Travis build: Break out sonar

matthieun:
The sonar command within the build.sh script would have masked build errors within cron jobs. Breaking it out to follow the same pattern as in atlas.


MikeGost:
@matthieun - while you're here - can you remove the protobuf dependency in `build.gradle` :)

matthieun:
Good idea!

PR #49, opened 2018-08-20
Added command line switch that allows saving to proto format

lucaspcram:
An optional command line switch `useProtoFormat=true` can be provided to generate atlas files using the protocol buffer format. When this switch is not provided, it will be set to false by default.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/persistence/AtlasProtoOutputFormat.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/persistence/AtlasProtoOutputFormat.java
@@ -1,0 +1,31 @@
+package org.openstreetmap.atlas.generator.persistence;
+
+import org.openstreetmap.atlas.geography.atlas.Atlas;
+import org.openstreetmap.atlas.geography.atlas.packed.PackedAtlas;
+import org.openstreetmap.atlas.geography.atlas.packed.PackedAtlas.AtlasSerializationFormat;
+import org.openstreetmap.atlas.streaming.resource.AbstractWritableResource;
+import org.openstreetmap.atlas.streaming.resource.FileSuffix;
+
+public class AtlasProtoOutputFormat extends AbstractFileOutputFormat<Atlas>
+{
+
+    @Override
+    protected String fileExtension()
+    {
+        return FileSuffix.ATLAS.toString();
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/generator/persistence/AtlasProtoOutputFormat.java
```diff
@@ -0,0 +1,37 @@
...
+import org.openstreetmap.atlas.streaming.resource.FileSuffix;
+
+/**
+ * {@link FileOutputFormat} that writes an {@link Atlas} in protocol buffer format.
+ * 
+ * @author lcram
+ */
+public class AtlasProtoOutputFormat extends AbstractFileOutputFormat<Atlas>
+{
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/generator/persistence/MultipleAtlasProtoOutputFormat.java
```diff
@@ -0,0 +1,29 @@
...
+    protected RecordWriter<String, Atlas> getBaseRecordWriter(final FileSystem fileSystem,
+            final JobConf job, final String name, final Progressable progress) throws IOException
+    {
+        if (this.format == null)
+        {
+            this.format = new AtlasProtoOutputFormat();
+        }
+        return this.format.getRecordWriter(fileSystem, job, name, progress);
+    }
+
```
MikeGost: Nit: extra line

lucaspcram: Good catch thanks!

lucaspcram:
I changed the switch so the generation is proto by default. Now you must provide a switch `-useJavaFormat=true` to use the old java format.

lucaspcram:
@matthieun I would imagine that the best way to unit test this would be to kick off a local Spark job, create a single atlas file (maybe we need a test PBF file in this repo?), and then verify its format using `atlas`. Would require some changes to `atlas` to allow for retrieval of serialization type.

Or are you thinking of something simpler that could fit in a unit test?

matthieun:
Added https://github.com/osmlab/atlas-generator/issues/51

PR #53, opened 2018-09-18
Added a RawAtlasCreator command

lucaspcram:
### Description:

NOTE: This is a WIP tool, and this PR represents iteration 1. I just needed to get something basic working to start tackling some other bugs.

The `RawAtlasCreator` is a new iteration of the `AtlasCreator` designed specifically for the raw atlas flow. When provided with the necessary OSM PBFs, it handles all the `DynamicAtlas` expansion underlying the `WaySectionProcessor` by building the necessary sliced atlases on the fly. It then stores them in a user specified location for re-use.

If you choose not to supply specific underlying OSM PBFs, then the build process will skip expansion into those missing shards.

You can also stop the generation process at any step in the raw atlas flow. This idea is encapsulated by the `RawAtlasFlavor` enum.

Ideas for future refactoring:

1. Allow for a more flexible PBF scheme, currently all files must be stored locally under a single folder.

2. Utilize the `ConcurrentResourceCache` instead of my ad hoc caching code.

3. General cleanup/commenting for ease of use/modification.

### Potential Impact:

No downstream impact. This is a standalone command.

### Unit Test Approach:

N/A

### Test Results:

I have tested the command locally to build various atlases, and tested them against known golden images. It seems to be working as expected.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/creator/RawAtlasCreator.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/creator/RawAtlasCreator.java
@@ -82,11 +82,22 @@ public class RawAtlasCreator extends Command
 
     private static final Logger logger = LoggerFactory.getLogger(RawAtlasCreator.class);
 
+    /*
+     * A path to a country boundary map file
+     */
     public static final Switch<CountryBoundaryMap> BOUNDARIES = new Switch<>("boundaries",
             "The boundary map to use", value -> CountryBoundaryMap.fromPlainText(new File(value)),
             Optionality.REQUIRED);
+
+    /*
+     * The ISO-3 country code of the country's shard to build
+     */
     public static final Switch<String> COUNTRY = new Switch<>("country", "The country code",
             StringConverter.IDENTITY, Optionality.REQUIRED);
+
+    /*
+     * The path to where the final output will be saved
+     */
     public static final Switch<File> OUTPUT = new Switch<>("output",
             "The path where the output will be saved", value ->
             {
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/generator/creator/RawAtlasCreator.java
```diff
@@ -0,0 +1,375 @@
...
+                logger.warn("Sliced cache miss for {}, generating...", atlasName);
+
+                try
+                {
+                    atlasOption = Optional.ofNullable(creator.generateSlicedAtlasFromScratch(
+                            pbfPath, countryName, countryBoundaryMap, shard));
+                }
+                catch (final Exception exception)
+                {
+                    atlasOption = Optional.empty();
```
MikeGost: We should probably log the exception here - would assume if there is a valid PBF path and we fail to generate the sliced atlas file for it - we have an issue.

src/main/java/org/openstreetmap/atlas/generator/creator/RawAtlasCreator.java
```diff
@@ -0,0 +1,375 @@
...
+ * @author lcram
+ */
+public class RawAtlasCreator extends Command
+{
+    /**
+     * Represents the different possible stages (or flavors) in the raw atlas flow.
+     *
+     * @author lcram
+     */
+    private enum RawAtlasFlavor
```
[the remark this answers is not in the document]
lucaspcram: I like this idea for the future. I'm going to leave It where it is for now, but I've made a note of this for a future update.

It would be neat to see some examples of places where you feel this enum could be used, other than just in this class!

src/main/java/org/openstreetmap/atlas/generator/creator/RawAtlasCreator.java
lucaspcram: Done.

src/main/java/org/openstreetmap/atlas/generator/creator/RawAtlasCreator.java
```diff
@@ -0,0 +1,378 @@
...
+        {
+            return this.flavorString;
+        }
+    }
+
+    private static final Logger logger = LoggerFactory.getLogger(RawAtlasCreator.class);
+    private static final String DEFAULT_CACHE_NAME = "__RawAtlasCreator_slicedAtlasCache__";
+
+    /*
+     * A path to a country boundary map file
```
matthieun: The javadoc-style comments are usually reserved for methods. For consistency, I'd use `//` here instead.

matthieun: Had not noticed all those are one-star... All good :)

PR #54, opened 2018-10-08
Switched caching to use HadoopAtlasFileCache instead of custom code

adahn6:
### Description:

This PR replaces the custom caching code for atlas files with HadoopAtlasFileCache. 

### Potential Impact:

In theory, nothing-- the existing code used the same approach to caching that the HadoopAtlasFileCache implementation does. Given that the implementations are slightly different, it is always possible that some performance changes could occur, but none were apparent during sanity tests.

### Unit Test Approach:

There does not seem to be a very good way to unit test this functionality; I am open to suggestions.

### Test Results:

All existing test pass.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
@@ -329,8 +329,6 @@ public final class AtlasGeneratorHelper implements Serializable
                 // Create the fetcher
                 final Function<Shard, Optional<Atlas>> slicedRawAtlasFetcher = AtlasGeneratorHelper
                         .atlasFetcher(atlasCache, country, possibleShards);
-
-                logger.info("Started sectioning raw Atlas for {}", countryShardString);
                 // Section the Atlas
                 atlas = new WaySectionProcessor(countryShard.getShard(), atlasLoadingOption,
                         sharding, slicedRawAtlasFetcher).run();
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
@@ -1,12 +1,7 @@
 package org.openstreetmap.atlas.generator;
 
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.Serializable;
-import java.nio.file.FileAlreadyExistsException;
-import java.nio.file.Files;
-import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -20,11 +15,9 @@ import org.apache.hadoop.fs.Path;
 import org.apache.spark.api.java.function.PairFlatMapFunction;
 import org.apache.spark.api.java.function.PairFunction;
 import org.openstreetmap.atlas.exception.CoreException;
-import org.openstreetmap.atlas.exception.ExceptionSearch;
 import org.openstreetmap.atlas.generator.persistence.scheme.SlippyTilePersistenceScheme;
+import org.openstreetmap.atlas.generator.tools.caching.HadoopAtlasFileCache;
 import org.openstreetmap.atlas.generator.tools.filesystem.FileSystemCreator;
-import org.openstreetmap.atlas.generator.tools.filesystem.FileSystemHelper;
-import org.openstreetmap.atlas.generator.tools.spark.utilities.SparkFileHelper;
 import org.openstreetmap.atlas.geography.atlas.Atlas;
 import org.openstreetmap.atlas.geography.atlas.AtlasResourceLoader;
 import org.openstreetmap.atlas.geography.atlas.delta.AtlasDelta;
@@ -37,7 +30,6 @@ import org.openstreetmap.atlas.geography.boundary.CountryBoundaryMap;
 import org.openstreetmap.atlas.geography.sharding.CountryShard;
 import org.openstreetmap.atlas.geography.sharding.Shard;
 import org.openstreetmap.atlas.geography.sharding.Sharding;
-import org.openstreetmap.atlas.streaming.resource.File;
 import org.openstreetmap.atlas.streaming.resource.FileSuffix;
 import org.openstreetmap.atlas.streaming.resource.Resource;
 import org.openstreetmap.atlas.streaming.resource.StringResource;
@@ -82,9 +74,6 @@ public final class AtlasGeneratorHelper implements Serializable
     /**
      * @param atlasDirectory
      *            The path of the folder containing the Atlas files, in format CTRY_z-x-y.atlas.gz
-     * @param temporaryDirectory
-     *            The path of the temporary folder to download Atlas files if they are not
-     *            downloaded already
      * @param country
      *            The country to look for
      * @param sparkContext
@@ -94,8 +83,8 @@ public final class AtlasGeneratorHelper implements Serializable
      * @return A function that returns an {@link Atlas} given a {@link Shard}
      */
     protected static Function<Shard, Optional<Atlas>> atlasFetcher(final String atlasDirectory,
-            final String temporaryDirectory, final String country,
-            final Map<String, String> sparkContext, final Set<Shard> validShards)
+            final String country, final Map<String, String> sparkContext,
+            final Set<Shard> validShards)
     {
         // & Serializable is very important as that function will be passed around by Spark, and
         // functions are not serializable by default.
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -106,115 +95,9 @@ public static ConfiguredTaggableFilter getTaggableFilterFrom(
...
-                final String path = SparkFileHelper.combine(atlasDirectory,
-                        String.format("%s%s", getAtlasName(country, shard), ATLAS_EXTENSION));
-                final Resource fileFromNetwork = FileSystemHelper.resource(path, sparkContext);
-                try
-                {
-                    return loadAtlas(fileFromNetwork);
-                }
-                catch (final Exception e)
-                {
-                    final Optional<FileNotFoundException> fileNotFound = ExceptionSearch
```
MikeGost: Do we no longer hit this case or is it handled by the cache itself?

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -106,115 +95,9 @@ public static ConfiguredTaggableFilter getTaggableFilterFrom(
...
-                        return Optional.empty();
-                    }
-                    else
-                    {
-                        // There is something else going on, re-throw and continue
-                        throw e;
-                    }
-                }
-            }
+            final HadoopAtlasFileCache atlasCache = new HadoopAtlasFileCache(atlasDirectory,
```
MikeGost: Is it correct understanding that each time the fetcher function is invoked, it will create a new `HadoopAtlasFileCache` instance or because we re-use the fetcher for all slicing operations, we are referencing the same underlying cache each time? I guess what I'm asking is, is it possible to persist the cache outside of the Function itself or does it not matter.

lucaspcram: So in @adahn6 's usage here, it will construct a new cache object (and new caching strategy object) each time. However, since we are always referencing the same underlying filesystem (in the case of a `SystemTemporaryFileCachingStrategy`), and because URIs are mapped to their MD5 sum, the actual cache contents persist across instances of any given `HadoopAtlasFileCache`. The only downside to recycling the object is that you lose the map of precomputed MD5 sums each recycle. So there will be a small performance penalty associated with recomputing the MD5 sum of a URI in order to look it up in the filesystem. However, you will still get cache hits across objects.

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
lucaspcram: The cache handles this. See my comment below.

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -106,115 +95,9 @@ public static ConfiguredTaggableFilter getTaggableFilterFrom(
...
+                    sparkContext);
+            return Optional.ofNullable(ATLAS_LOADER.load(atlasCache.get(country, shard).get()));
```
lucaspcram: I would suggest modifying this slightly, since it is possible for `atlasCache.get(country, shard)` to return an empty optional. In that case, the unwrapping `get` call would pose a problem, and we would want to bubble up an empty optional to the caller. Maybe something like:

```java
Optional<Resource> atlasResource = atlasCache.get(country, shard);
if (atlasResource.isPresent()) {
    return Optional.ofNullable(ATLAS_LOADER.load(atlasResource.get()));
} else {
    return Optional.empty()l;
}
```

lucaspcram:
Looks good @adahn6 , this really cleans up the generator code. Just one minor thing that may need some updating.

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
MikeGost: Makes sense, this may be worth a comment, as it's not obvious by just reading the code. Or perhaps we could pull it out of the function itself, so it's more obvious, your call @adahn6!

lucaspcram: I am partial to pulling it out into a `static final` class variable, like how we have the `ATLAS_LOADER` class variable for the `AtlasResourceLoader`. `HadoopAtlasFileCache` is threadsafe so it should not be a problem.

adahn6: didn't realize it was thread safe! I instantiated a new object since that seemed safest, but I will pull it out into a `static final` class variable :)

adahn6: wait, I remember why I didn't do that now, several of the function parameters are required for the HadoopAtlasFileCache constrctor (spark context and atlas directory). because of that, there's not a good way to make them `static final`. I could, however, make one cache object and pass it in to the function?

MikeGost: That sounds reasonable.

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -444,14 +323,14 @@ protected static AtlasLoadingOption buildAtlasLoadingOption(final CountryBoundar
...
+                final HadoopAtlasFileCache atlasCache = new HadoopAtlasFileCache(slicedRawAtlasPath,
+                        sparkContext);
                 // Create the fetcher
                 final Function<Shard, Optional<Atlas>> slicedRawAtlasFetcher = AtlasGeneratorHelper
-                        .atlasFetcher(SparkFileHelper.combine(slicedRawAtlasPath, country),
-                                System.getProperty("java.io.tmpdir"), country, sparkContext,
-                                possibleShards);
+                        .atlasFetcher(atlasCache, country, possibleShards);
 
+                logger.info("Started sectioning raw Atlas for {}", countryShardString);
```
MikeGost: Nit: Sorry, you'll hate me for this - can you remove this please? The WaySectionProcessor has very fine-grained logging already and logs this exact line at it's level.

adahn6: np, sane logging is important!

adahn6:
sanity test of PR updates passed :)

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
lucaspcram: Ah shoot that's right. Originally, I designed the cache classes (`ConcurrentResourceCache` and `HadoopAtlasFileCache`) to allow you to change the parameters with some `set` methods. But then this compromised thread safety because there would be a race between the setters and the caches' `get` methods (since a `get` could preempt one of the `set` calls before it can properly complete).

One way to get around this would be to have the constructor take a lock object, which the user would have to pass in themselves. You could then lock the `set` and `get` methods with that common lock. I decided against this since it makes the class more complicated. However, as these classes are integrated into our codebase, maybe it makes more sense to change the design?

adahn6: changing the directory sounds like it would make things really complicated for the MD5 map-- without updating it too, the mapping is probably invalidated, and updating that map wouldn't necessarily be fast or easy. IMO, I think the current implementation is good for now. As use cases expand, we can examine more complex projects with multiple working directories to see what's best. 
Off the top of my head, I'd consider a `CacheManager` that takes requests for caches and gives back an instantiated object; abstracted in that class would be logic around "if a cache exists, return it else make a new one for that directory".

adahn6: wait nm, I was misreading some code and thought it was just hashing based off filename

lucaspcram: It's based off the full URI. You're not the first person to mix that up :)

adahn6: could we just switch the cache to use URIs then? seems like a pretty minor change overall, but allows for the reuse of one cache object everywhere

adahn6: if we wanted to keep filename to URI conversion logic in the cache itself, could make it a static method that takes in filename, base dir, and spark config and returns the expected URI

PR #59, opened 2018-10-30
DNM: AtlasGenerator cleanup and fix AtlasScheme being dropped

matthieun:
### Description:

This PR removes references to the older Atlas generation straight from PBF. It now defaults to the RawAtlas, and the option to roll back to the older one disappears.
Also:
- Create `AtlasGeneratorParameters` to reduce complexity. This one does just Switch definition and configuration building.
- Enable Atlas scheme (the configurable file path with which Atlas files are saved) which did not work with RawAtlas. It did only work with an empty scheme.
- Depend on Atlas 5.2.3, to benefit from the latest caching improvements: https://github.com/osmlab/atlas/pull/258

### Potential Impact:

Non Raw atlas builds will be impossible.

### Unit Test Approach:

- Used existing tests
- Fixed AtlasGeneratorIntegrationTest to use the raw atlas folder scheme
- Updated HadoopAtlasFileCache to be aware of the atlas folder scheme

### Test Results:

DNM until large scale test validates it works.

------

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,6 +1,6 @@
 project.ext.versions = [
     checkstyle: '7.6.1',
-    atlas: '5.1.8',
+    atlas: '5.2.3',
     spark: '1.6.0-cdh5.7.0',
     snappy: '1.1.1.6',
 ]
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
@@ -11,6 +11,7 @@ import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapred.JobConf;
 import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.broadcast.Broadcast;
+import org.openstreetmap.atlas.generator.AtlasGeneratorHelper.NamedAtlasStatistics;
 import org.openstreetmap.atlas.generator.persistence.MultipleAtlasCountryStatisticsOutputFormat;
 import org.openstreetmap.atlas.generator.persistence.MultipleAtlasOutputFormat;
 import org.openstreetmap.atlas.generator.persistence.MultipleAtlasProtoOutputFormat;
@@ -254,8 +255,12 @@ public class AtlasGenerator extends SparkJob
                 {
                     final String countryShardName = tuple._1();
                     final String countryName = StringList.split(countryShardName, "_").get(0);
-                    return new Tuple2<>(countryName, tuple._2());
-                }).reduceByKey(AtlasStatistics::merge);
+                    return new Tuple2<>(countryName,
+                            // Here using NamedAtlasStatistics so the reduceByKey function below can
+                            // name what statistic merging failed, if any.
+                            new NamedAtlasStatistics(countryName, tuple._2()));
+                }).reduceByKey(AtlasGeneratorHelper.reduceAtlasStatistics())
+                .mapToPair(tuple -> new Tuple2<>(tuple._1(), tuple._2().getAtlasStatistics()));
 
         // Save aggregated metrics
         this.getContext().setJobGroup("4", "Country Statistics Creation");
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
@@ -125,8 +125,7 @@ public class AtlasGenerator extends SparkJob
         final String shardingName = (String) command.get(AtlasGeneratorParameters.SHARDING_TYPE);
         final Sharding sharding = AtlasSharding.forString(shardingName, configuration());
         final Sharding pbfSharding = pbfShardingName != null
-                ? AtlasSharding.forString(pbfShardingName, configuration())
-                : sharding;
+                ? AtlasSharding.forString(pbfShardingName, configuration()) : sharding;
         final PbfContext pbfContext = new PbfContext(pbfPath, pbfSharding, pbfScheme);
         final String shouldAlwaysSliceConfiguration = (String) command
                 .get(AtlasGeneratorParameters.SHOULD_ALWAYS_SLICE_CONFIGURATION);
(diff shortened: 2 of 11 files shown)


dependencies.gradle
```diff
@@ -1,6 +1,6 @@
 project.ext.versions = [
     checkstyle: '7.6.1',
-    atlas: '5.1.8',
+    atlas: '5.2.3',
```
lucaspcram: I think you might be looking for `atlas 5.2.2`, no?

matthieun: Nope, waiting for 5.2.3 :)

lucaspcram: Oh, nevermind. I understand what you are doing now :)

src/main/java/org/openstreetmap/atlas/generator/tools/caching/HadoopAtlasFileCache.java
```diff
@@ -55,9 +61,14 @@ public HadoopAtlasFileCache(final String parentAtlasPath,
...
+        String compiledAtlasScheme = "";
+        if (shard instanceof SlippyTile)
+        {
+            compiledAtlasScheme = this.atlasScheme.compile((SlippyTile) shard);
+        }
         final String atlasName = String.format("%s_%s", country, shard.getName());
-        final String atlasURIString = this.parentAtlasPath + "/" + country + "/" + atlasName
-                + FileSuffix.ATLAS.toString();
+        final String atlasURIString = this.parentAtlasPath + "/" + country + "/"
+                + compiledAtlasScheme + atlasName + FileSuffix.ATLAS.toString();
```
lucaspcram: My plan for the next update was to change all the `+ "/" +` constructs to `SparkFileHelper.combine()` calls. Since it looks like `compiledAtlasScheme` already has a trailing `/`, we'll have to figure out what to do here!

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorParameters.java
```diff
@@ -0,0 +1,192 @@
...
+import org.openstreetmap.atlas.utilities.conversion.StringConverter;
+import org.openstreetmap.atlas.utilities.runtime.Command.Optionality;
+import org.openstreetmap.atlas.utilities.runtime.Command.Switch;
+import org.openstreetmap.atlas.utilities.runtime.Command.SwitchList;
+import org.openstreetmap.atlas.utilities.runtime.CommandMap;
+
+/**
+ * @author matthieun
+ */
+public final class AtlasGeneratorParameters
```
lucaspcram: Good idea.

src/main/java/org/openstreetmap/atlas/generator/tools/caching/HadoopAtlasFileCache.java
```diff
@@ -27,21 +29,25 @@
...
      *            The configuration map
      */
     public HadoopAtlasFileCache(final String parentAtlasPath,
-            final Map<String, String> configuration)
+            final SlippyTilePersistenceScheme atlasScheme, final Map<String, String> configuration)
     {
         super(new SystemTemporaryFileCachingStrategy(),
                 uri -> FileSystemHelper.resource(uri.toString(), configuration));
         this.parentAtlasPath = parentAtlasPath;
+        this.atlasScheme = atlasScheme;
```
lucaspcram: What happens if the user passes a `null` scheme, or does not want to pass in a scheme? It is possible downstream users may not anticipate this constructor change, or may not necessarily need this functionality. Perhaps we should overload the constructor and have a sensible default scheme?

I am planning on adding additional constructors to support the new `NamespaceCachingStrategy`, so this is something I may be able to do in a later PR.

src/main/java/org/openstreetmap/atlas/generator/tools/caching/HadoopAtlasFileCache.java
matthieun: Maybe skip the `SparkFileHelper.combine()`?

src/main/java/org/openstreetmap/atlas/generator/tools/caching/HadoopAtlasFileCache.java
matthieun: Yes, we can make it use defaults... Let me do that

src/main/java/org/openstreetmap/atlas/generator/tools/caching/HadoopAtlasFileCache.java
lucaspcram: I suppose so. I just always tend to recoil from hardcoding path separators. What if somebody wants to use Windows ;)

src/main/java/org/openstreetmap/atlas/generator/tools/caching/HadoopAtlasFileCache.java
matthieun: Done..

matthieun:
Closing to re-trigger

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -84,71 +61,6 @@ public String apply(final String key)
     public static final String SHARD_DELTAS_CHANGED_FOLDER = "deltasChanged";
     public static final String SHARD_DELTAS_REMOVED_FOLDER = "deltasRemoved";
 
-    private static final String SHARDING_DEFAULT = "slippy@10";
-
-    public static final Switch<StringList> COUNTRIES = new Switch<>("countries",
```
MikeGost: A lot cleaner, I'm a fan.

src/main/java/org/openstreetmap/atlas/generator/tools/caching/HadoopAtlasFileCache.java
```diff
@@ -39,9 +39,29 @@
     public HadoopAtlasFileCache(final String parentAtlasPath,
             final Map<String, String> configuration)
     {
-        super(new SystemTemporaryFileCachingStrategy(),
-                uri -> FileSystemHelper.resource(uri.toString(), configuration));
+        super(new NamespaceCachingStrategy("HadoopAtlasFileCache"), uri -> Optional
```
lucaspcram: We have a `GlobalNamespaceCachingStrategy` for this purpose!

PR #74, opened 2019-01-21
Refactored SlippyTilePersistenceScheme and other path code

lucaspcram:
### Description:

~DNM: doing some last minute testing. Hold merge until ready.~
Ready for merge.

Refactored the `SlippyTilePersistenceScheme` to disallow unsanitized user schemes. Additionally, all hardcoded path construction code (i.e. `foo + "/" + bar`) has been removed.

### Potential Impact:

Potential downstream impact. `SlippyTilePersistenceScheme` still recognizes parameters in formats like `zz/xx/yy`, but now it will throw errors if the format does not match a predefined scheme in `SlippyTilePersistenceSchemeType`. If downstream code is using an unrecognized scheme string, it will either need to be added as a new type in `SlippyTilePersistenceSchemeType`, or the downstream code must update.

### Unit Test Approach:

New unit tests added (see diff), and old unit/integration tests pass.

### Test Results:

All tests pass. Additionally, I built some atlas files locally and things look OK.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/persistence/scheme/SlippyTilePersistenceScheme.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/persistence/scheme/SlippyTilePersistenceScheme.java
@@ -3,8 +3,30 @@ package org.openstreetmap.atlas.generator.persistence.scheme;
 import java.io.Serializable;
 
 import org.openstreetmap.atlas.geography.sharding.SlippyTile;
+import org.openstreetmap.atlas.streaming.resource.FileSuffix;
 
 /**
+ * This class helps users construct storage paths dynamically based on the name of the target
+ * resource to be saved. The various allowed storage schemes are defined in
+ * {@link SlippyTilePersistenceSchemeType}.<br>
+ * <br>
+ * For example, one PBF storage scheme is 'ZZ_XX_YY_SUBFOLDERS_PBF', which is defined as
+ * 'zz/xx/yy/zz-xx-yy.pbf'. This means that when saving the PBF '12-34-48.pbf' to a remote store,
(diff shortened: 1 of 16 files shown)


src/main/java/org/openstreetmap/atlas/generator/tools/spark/SparkJob.java
```diff
@@ -425,8 +426,9 @@ private void writeStatus(final String path, final String name, final String cont
         try
         {
             final FileSystem fileSystem = getFileSystem(path);
-            final BufferedWriter out = new BufferedWriter(
-                    new OutputStreamWriter(fileSystem.create(new Path(path + "/" + name))));
+            final BufferedWriter out = new BufferedWriter(new OutputStreamWriter(
+                    fileSystem.create(new Path(SparkFileHelper.combine(path, name)))));
+            // TODO this leaks a resource if write() throws an exception
```
MikeGost: Can you put the `BufferedWriter` in a try-with-resources block to avoid the leak?

lucaspcram: I don't see why not.

src/main/java/org/openstreetmap/atlas/generator/persistence/scheme/SlippyTilePersistenceScheme.java
```diff
@@ -15,11 +37,34 @@
...
     private final String scheme;
+    private final FileSuffix suffix;
+
+    public static SlippyTilePersistenceScheme getSchemeInstanceFromString(final String string)
+    {
+        SlippyTilePersistenceSchemeType type;
+
+        /*
+         * Backwards compatibility hack: for now, allow this class to parse any old scheme strings
+         * still found in the wild. Ideally, downstream users should migrate calling code to use the
```
matthieun: This is wild.

PR #76, opened 2019-02-12
PbfLocator: Protect against file system failures

matthieun:
### Description:

Use retries when loading PBF files from a Hadoop `FileSystem`

### Potential Impact:

PBF loading in `AtlasGenerator` will be more robust.

### Unit Test Approach:

This is tested by the `AtlasGeneratorIntegrationTest`.

### Test Results:

`AtlasGeneratorIntegrationTest` passes.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/PbfLocator.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/PbfLocator.java
@@ -1,7 +1,8 @@
 package org.openstreetmap.atlas.generator;
 
 import java.io.IOException;
-import java.io.Serializable;
+// SONAR fails to see this is used to cast lambda functions.
+import java.io.Serializable; // NOSONAR
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
--- a/src/main/java/org/openstreetmap/atlas/generator/PbfLocator.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/PbfLocator.java
@@ -8,6 +8,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 
+import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.openstreetmap.atlas.exception.CoreException;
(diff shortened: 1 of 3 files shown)


matthieun:
Will re-open after some more thought...

src/main/java/org/openstreetmap/atlas/generator/PbfLocator.java
```diff
@@ -32,15 +36,13 @@
  *
  * @author matthieun
  */
-public class PbfLocator implements Serializable
```
lucaspcram: Why did we ditch `Serializable`?

src/main/java/org/openstreetmap/atlas/generator/PbfLocator.java
```diff
@@ -124,12 +119,19 @@ public PbfLocator(final PbfContext pbfContext, final Map<String, String> spark)
      */
     public Iterable<LocatedPbf> pbfsCovering(final MultiPolygon multiPolygon)
     {
-        logger.trace("Seeking tiles for MultiPolygon {}", multiPolygon.toSimpleString());
+        if (logger.isTraceEnabled())
```
lucaspcram: Is this a Sonar thing?

src/main/java/org/openstreetmap/atlas/generator/creator/RawAtlasCreator.java
```diff
@@ -263,7 +265,7 @@ protected SwitchList switches()
                 }
                 catch (final Exception exception)
                 {
-                    logger.error("{}", exception.toString());
```
lucaspcram: Good catch.

src/main/java/org/openstreetmap/atlas/generator/PbfLocator.java
matthieun: Yes. I realized Sonar complained that some of its members were not serializable, and not transient, which means we never used it, otherwise it would have failed.

src/main/java/org/openstreetmap/atlas/generator/PbfLocator.java
matthieun: Yes, to avoid calling `multiPolygon.toSimpleString()` if not necessary.

src/main/java/org/openstreetmap/atlas/generator/creator/RawAtlasCreator.java
adahn6: +1

src/main/java/org/openstreetmap/atlas/generator/PbfLocator.java
adahn6: good catch!

lucaspcram:
Should I merge in spite of this Sonar error (it is complaining about an unused `Serializable` import)

matthieun:
Yes, this one is a false positive. We are using that import in the lambda casts.

PR #81, opened 2019-02-22
Integration Test for dynamic relation slicing

jwpgage:
### Description:

This PR is an integration test for the recent dynamic relation slicing updates. This tests relation slicing a water relation across a country boundary, as well as across four shard boundaries.

The original water relation, with the red country boundary and the blue shading tree (note that the bottom two shards did not render in):

screen shot 2019-02-22 at 8 56 24 am

After relation slicing is run, here is the closed relation from the view of the top country (ABC):

screen shot 2019-02-22 at 8 57 03 am

And here is the view of the bottom country (DEF):

screen shot 2019-02-22 at 8 57 41 am

The areas of the resulting polygons are then verified.

### Potential Impact:

The access of `AtlasGeneratorHelper.atlasFetcher()` has been changed to public, to allow for ease of use.

### Unit Test Approach:

This PR is an integration test.

### Test Results:

Integration test passes!

------


src/integrationTest/java/org/openstreetmap/atlas/generator/slicing/DynamicRelationSlicingTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.openstreetmap.atlas.geography.sharding.DynamicTileSharding;
+import org.openstreetmap.atlas.geography.sharding.Shard;
+import org.openstreetmap.atlas.geography.sharding.Sharding;
+import org.openstreetmap.atlas.streaming.resource.File;
+import org.openstreetmap.atlas.streaming.resource.InputStreamResource;
+
+/**
+ * @author james-gage
+ */
+public class DynamicRelationSlicingTest
```
matthieun: If this is an integration test, I think we should call it `DynamicRelationSlicingIntegrationTest` to avoid confusion. There are other tests in the atlas repos that are not named this way and I find it confusing!

src/integrationTest/java/org/openstreetmap/atlas/generator/slicing/DynamicRelationSlicingTest.java
```diff
@@ -0,0 +1,118 @@
...
+        addResource(ABC_101, "ABC_1-0-1.atlas");
+        addResource(ABC_110, "ABC_1-1-0.atlas");
+        addResource(ABC_111, "ABC_1-1-1.atlas");
+        addResource(DEF_101, "DEF_1-0-1.atlas");
+        addResource(DEF_111, "DEF_1-1-1.atlas");
+    }
+
+    private static void addResource(final String path, final String name)
+    {
+        ResourceFileSystem.addResource(path, new InputStreamResource(
```
matthieun: Maybe store the Atlas files as text (to avoid binary) and then load them into `PackedAtlas` here

src/integrationTest/java/org/openstreetmap/atlas/generator/slicing/DynamicRelationSlicingTest.java
```diff
@@ -0,0 +1,118 @@
...
+     * This integration test country slices a water relation that spans four shards, and crosses a
+     * country boundary. The relation is country sliced twice, once from the ABC perspective and
+     * once from DEF. The resulting closed relations then have their areas validated.
+     */
+    @Test
+    public void testRelationSlicing()
+    {
+        // parameter setup
+        final Sharding sharding = new DynamicTileSharding(new File(
+                DynamicRelationSlicingTest.class.getResource("simpleSharding.txt").getFile()));
```
matthieun: Why not use the same pattern as below for this one?

src/integrationTest/java/org/openstreetmap/atlas/generator/slicing/DynamicRelationSlicingTest.java
```diff
@@ -0,0 +1,118 @@
...
+                        .getResourceAsStream("simpleBoundaryFile.txt")));
+
+        final Map<String, String> sparkOptions = new HashMap<>();
+        sparkOptions.put("spark.executor.memory", "512m");
+        sparkOptions.put("spark.executor.memory", "512m");
+        sparkOptions.put("spark.driver.memory", "512m");
+        ResourceFileSystem.simpleconfiguration()
+                .forEach((key, value) -> sparkOptions.put(key, value));
+
+        final HadoopAtlasFileCache lineSlicedAtlasCache = new HadoopAtlasFileCache(
```
matthieun: Nice mix of the cache and the resource file system!

src/integrationTest/java/org/openstreetmap/atlas/generator/slicing/DynamicRelationSlicingTest.java
```diff
@@ -0,0 +1,118 @@
...
+        double totalAreaDEF = 0;
+        for (final Line line : relationSlicedAtlasDEF.lines())
+        {
+            if (line.isClosed())
+            {
+                final Polygon polygon = new Polygon(line.asPolyLine());
+                totalAreaDEF += polygon.surface().asDm7Squared();
+            }
+        }
+        Assert.assertTrue(totalAreaDEF == 5.9870426059188685E17);
```
matthieun: Maybe use assertEquals here so if it diverges in the future it will tell what it saw and that it is different?

PR #86, opened 2019-03-18
Streamline AtlasGenerator and improve slicing performance

adahn6:
### Description:

This PR reworks AtlasGenerator to use a new enum, AtlasGeneratorJobGroup, to manage the Strings and output formats associated with each stage of the Spark job for AtlasGenerator. Additionally, the performance of relation slicing is improved-- now, the lineSlicedSubAtlas creation will use a more aggressive filter that keeps only lines and relations tagged with appropriate tags, and uses SilkCut to ensure their underlying Points are also included. This dramatically reduces the amount of geometry loaded by the dynamic exploration during relation slicing. 

### Potential Impact:

None, as all data should be the same.

### Unit Test Approach:

None, as all data should be the same.

### Test Results:

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
@@ -9,7 +9,6 @@ import java.util.stream.Collectors;
 
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapred.JobConf;
-import org.apache.hadoop.mapred.lib.MultipleOutputFormat;
 import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.broadcast.Broadcast;
 import org.openstreetmap.atlas.generator.AtlasGeneratorHelper.NamedAtlasStatistics;
@@ -180,7 +179,8 @@ public class AtlasGenerator extends SparkJob
                 .parallelize(tasks, tasks.size())
                 .mapToPair(AtlasGeneratorHelper.generateRawAtlas(broadcastBoundaries, sparkContext,
                         broadcastLoadingOptions, pbfContext, atlasScheme))
-                .filter(tuple -> tuple._2() != null).cache();
+                .filter(tuple -> tuple._2() != null);
+        countryRawAtlasRDD.cache();
 
         // Persist the RDD and save the intermediary state
         saveAsHadoop(countryRawAtlasRDD, AtlasGeneratorJobGroup.RAW, output);
@@ -188,7 +188,8 @@ public class AtlasGenerator extends SparkJob
         // Slice the raw Atlas and filter any null atlases
         final JavaPairRDD<String, Atlas> lineSlicedAtlasRDD = countryRawAtlasRDD
                 .mapToPair(AtlasGeneratorHelper.sliceRawAtlasLines(broadcastBoundaries))
-                .filter(tuple -> tuple._2() != null).cache();
+                .filter(tuple -> tuple._2() != null);
+        lineSlicedAtlasRDD.cache();
         saveAsHadoop(lineSlicedAtlasRDD, AtlasGeneratorJobGroup.LINE_SLICED, output);
 
         // Remove the raw atlas RDD from cache since we've cached the sliced RDD
@@ -198,7 +199,8 @@ public class AtlasGenerator extends SparkJob
         final JavaPairRDD<String, Atlas> lineSlicedSubAtlasRDD = lineSlicedAtlasRDD
                 .mapToPair(AtlasGeneratorHelper.subatlas(AtlasGeneratorHelper.subAtlasFilter,
                         AtlasCutType.SILK_CUT))
-                .filter(tuple -> tuple._2() != null).cache();
+                .filter(tuple -> tuple._2() != null);
+        lineSlicedSubAtlasRDD.cache();
         saveAsHadoop(lineSlicedSubAtlasRDD, AtlasGeneratorJobGroup.LINE_SLICED_SUB, output);
 
         // Relation slice the line sliced Atlas and filter any null atlases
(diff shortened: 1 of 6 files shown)


src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -286,15 +249,8 @@ public void start(final CommandMap command)
...
-                .mapToPair(AtlasGeneratorHelper.generateAtlasStatistics(broadcastSharding));
-
-        // Persist the RDD and save
-        statisticsRDD.cache();
-        this.getContext().setJobGroup("5", "Shard Statistics Creation");
-        statisticsRDD.saveAsHadoopFile(getAlternateSubFolderOutput(output, SHARD_STATISTICS_FOLDER),
-                Text.class, AtlasStatistics.class, MultipleAtlasStatisticsOutputFormat.class,
-                new JobConf(configuration()));
-        logger.info("\n\n********** SAVED THE SHARD STATISTICS **********\n");
+                .mapToPair(AtlasGeneratorHelper.generateAtlasStatistics(broadcastSharding)).cache();
```
matthieun: How about we keep the cache in a separate line? It seems hidden there!

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -190,87 +180,60 @@ public void start(final CommandMap command)
...
-        // Section the sliced raw Atlas
+        // Section the sliced Atlas
         final JavaPairRDD<String, Atlas> countryAtlasShardsRDD = fullySlicedRawAtlasShardsRDD
                 .mapToPair(AtlasGeneratorHelper.sectionRawAtlas(broadcastBoundaries,
                         broadcastSharding, sparkContext, broadcastLoadingOptions,
-                        fullySlicedRawAtlasPath, atlasScheme, tasks));
+                        getAlternateSubFolderOutput(output,
+                                AtlasGeneratorJobGroup.FULLY_SLICED.getCacheFolder()),
+                        atlasScheme, tasks))
+                .cache();
```
matthieun: Same here, have the cache as a separate statement?

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -354,4 +299,17 @@ protected SwitchList switches()
...
     }
+
+    @SuppressWarnings("unchecked")
+    private void saveAsHadoop(final JavaPairRDD<?, ?> atlasRDD, final AtlasGeneratorJobGroup group,
+            final String output)
+    {
+        this.getContext().setJobGroup(group.getId().toString(), group.getDescription());
+        atlasRDD.saveAsHadoopFile(getAlternateSubFolderOutput(output, group.getCacheFolder()),
+                Text.class, group.getKeyClass(),
+                (Class<? extends MultipleOutputFormat<?, ?>>) group.getOutputClass(),
```
matthieun: So many questions... 🤔

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorJobGroup.java
```diff
@@ -0,0 +1,110 @@
...
+            "fullySlicedAtlas",
+            Atlas.class,
+            MultipleAtlasOutputFormat.class),
+    WAY_SECTIONED(
+            4,
+            "Way Sectioned Atlas Creation",
+            "atlas",
+            Atlas.class,
+            MultipleAtlasOutputFormat.class),
+    WAY_SECTIONED_PBF(
```
matthieun: Is this one a duplicate?

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -190,87 +180,60 @@ public void start(final CommandMap command)
                 .parallelize(tasks, tasks.size())
                 .mapToPair(AtlasGeneratorHelper.generateRawAtlas(broadcastBoundaries, sparkContext,
                         broadcastLoadingOptions, pbfContext, atlasScheme))
-                .filter(tuple -> tuple._2() != null);
+                .filter(tuple -> tuple._2() != null).cache();
```
matthieun: Oh and same here about having the cache on a separate statement.

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorJobGroup.java
adahn6: different output format! MultipleAtlasOutputFormat vs MultipleAtlasProtoOutputFormat

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
adahn6: this block of code was repeated throughout `start()`, but with different constants being applied. the constants all varied based on the task being done, so while this is a much more generic block of code, I think it's cleaner -- now, to figure out what each task is doing, we're managing those constants explicitly instead of just hard coding them into the flow

adahn6: unrelated but also this is slightly out of date, I will update to fix the unchecked and casting here

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
adahn6: changed!

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorJobGroup.java
matthieun: ooooh!

PR #88, opened 2019-03-19
Added in configuration option for filtered output

adahn6:
### Description:

Adds an option to create a separate, filtered output after final Atlas generation is finished. If no filter is provided, the filtering step will not be run. If a filter is provided, it should be in the form of a JSON taggable filter file; after Atlas generation, all generated Atlas files will be written to the 'filteredOutput' subdirectory using a SOFT_CUT subatlas call. 

### Potential Impact:
None, as there are no changes if the filter is not provided.

### Unit Test Approach:

Describe unit tests being added with that PR.

### Test Results:

------


src/integrationTest/java/org/openstreetmap/atlas/generator/AtlasGeneratorIntegrationTest.java
```diff
@@ -83,11 +83,11 @@ public void testAtlasGeneration()
                     new Path(LINE_DELIMITED_GEOJSON_OUTPUT + "/9/DMA_9-168-233.ldgeojson.gz")));
             Assert.assertTrue(resourceFileSystem.exists(
                     new Path(LINE_DELIMITED_GEOJSON_OUTPUT + "/9/DMA_9-168-234.ldgeojson.gz")));
-            Assert.assertEquals(402,
+            Assert.assertEquals(405,
```
matthieun: I believe this test fails in Travis mainline builds (Skipped in PRs)

PR #103, opened 2019-06-11
Tooling to save Sharding and Boundaries to output folder

matthieun:
### Description:

This allows the `AtlasGenerator` job to save the Sharding tree and Boundary file to the output folder. If present in the input PBF folder, it also uses those sharding and boundary files in priority, rather than the ones supplied in the parameters.

### Potential Impact:

This should be transparent to existing toolchains, as the existing parameters are still valid.

### Unit Test Approach:

Updated `AtlasGeneratorIntegrationTest`

### Test Results:

Tests pass. Also the `./gradlew run` command still works as expected.

------

### DIFF
--- a/src/integrationTest/java/org/openstreetmap/atlas/generator/AtlasGeneratorIntegrationTest.java
+++ b/src/integrationTest/java/org/openstreetmap/atlas/generator/AtlasGeneratorIntegrationTest.java
@@ -6,11 +6,13 @@ import org.apache.hadoop.fs.Path;
 import org.junit.Assert;
 import org.junit.Test;
 import org.openstreetmap.atlas.exception.CoreException;
+import org.openstreetmap.atlas.generator.tools.spark.persistence.PersistenceTools;
 import org.openstreetmap.atlas.generator.tools.streaming.ResourceFileSystem;
 import org.openstreetmap.atlas.streaming.compression.Decompressor;
 import org.openstreetmap.atlas.streaming.resource.FileSuffix;
 import org.openstreetmap.atlas.streaming.resource.InputStreamResource;
 import org.openstreetmap.atlas.streaming.resource.Resource;
+import org.openstreetmap.atlas.streaming.resource.StringResource;
 import org.openstreetmap.atlas.utilities.collections.Iterables;
 import org.openstreetmap.atlas.utilities.collections.StringList;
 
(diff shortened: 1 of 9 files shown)


src/integrationTest/java/org/openstreetmap/atlas/generator/AtlasGeneratorIntegrationTest.java
```diff
@@ -87,10 +94,19 @@ public void testAtlasGeneration()
                     Iterables.size(resourceForName(resourceFileSystem,
                             LINE_DELIMITED_GEOJSON_OUTPUT + "/9/DMA_9-168-234.ldgeojson.gz")
                                     .lines()));
-            Assert.assertEquals(402,
+            Assert.assertEquals(336,
```
adahn6: any reason for this change?

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -72,7 +72,7 @@ public static void main(final String[] args)
      *            {@link Sharding} strategy
      * @return {@link List} of {@link AtlasGenerationTask}s
      */
-    protected static List<AtlasGenerationTask> generateTasks(final StringList countries,
+    protected static List<AtlasGenerationTask> generateTasks(final List<String> countries,
```
adahn6: 👍

src/integrationTest/java/org/openstreetmap/atlas/generator/AtlasGeneratorIntegrationTest.java
matthieun: yes. the method below `resourceForName` had the shard 233 path hardcoded by mistake. Now it really loads the one asked for, and as a result, the 234 numbers are different....

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -270,15 +238,9 @@ public void start(final CommandMap command)
...
                         atlasScheme, tasks));
         countryAtlasShardsRDD.cache();
-
-        if (useJavaFormat)
-        {
-            saveAsHadoop(countryAtlasShardsRDD, AtlasGeneratorJobGroup.WAY_SECTIONED, output);
-        }
-        else
-        {
-            saveAsHadoop(countryAtlasShardsRDD, AtlasGeneratorJobGroup.WAY_SECTIONED_PBF, output);
```
adahn6: oh good-- can we remove the `AtlasGeneratorJobGroup.WAY_SECTIONED` enum value too?

src/integrationTest/java/org/openstreetmap/atlas/generator/AtlasGeneratorIntegrationTest.java
adahn6: that makes sense!

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
matthieun: Oh yes, good catch

src/main/java/org/openstreetmap/atlas/generator/tools/spark/persistence/PersistenceTools.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public Sharding sharding(final String input)
+    {
+        final Configuration hadoopConfiguration = hadoopConfiguration();
+        final Path inputPath = new Path(strip(input) + SHARDING_FILE);
+        return AtlasSharding.forString("dynamic@" + inputPath.toUri().toString(),
+                hadoopConfiguration);
+    }
+
+    final String strip(final String input)
```
lucaspcram: Maybe `appendDirectorySeparator` is a better name? Since this doesn't really strip anything away.

src/main/java/org/openstreetmap/atlas/generator/AtlasGenerator.java
```diff
@@ -398,4 +421,17 @@ private void saveAsHadoop(final JavaPairRDD<?, ?> atlasRDD, final AtlasGenerator
                 new JobConf(configuration()));
         logger.info(SAVED_MESSAGE, group.getDescription());
     }
+
+    private Sharding sharding(final String shardingName, final CommandMap command)
+    {
+        if (shardingName == null)
+        {
+            return persistenceTools()
+                    .sharding((String) command.get(AtlasGeneratorParameters.PBF_PATH));
```
jwpgage: Why do we default to the Pbf path here? It won't contain the sharding tree name, and the AtlasSharding.forString() on like 434 will always fail, no?

matthieun: So this method here is for the Atlas sharding. If the parameter is not provided, it will use the PersistenceTools to load it from the input PBF folder. Does that make sense?

matthieun:
Large scale test failed, closing until resolved.

matthieun:
False alarm, I made a mistake in the version. This works

PR #108, opened 2019-07-16
Added configured filter output option to Atlas Generation

jwpgage:
### Description:

This PR adds an option to Altas generation for configured filter output. This output is optional, and allows for more complicated predicates to be used in filtering output than the `ConfiguredTaggableFilter`.

Also added in this PR is the `WorldMultiAtlasGenerator`, which is a command that takes an atlas directory and creates a corresponding multi atlas.

### Potential Impact:

Adds an option for configured filtered output to atlas generation.

### Unit Test Approach:

Tested atlas generation with and without the optional configured filter.

### Test Results:

Filtered output was as expected.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
@@ -35,6 +35,7 @@ import org.openstreetmap.atlas.geography.sharding.Sharding;
 import org.openstreetmap.atlas.streaming.resource.Resource;
 import org.openstreetmap.atlas.tags.Taggable;
 import org.openstreetmap.atlas.utilities.collections.StringList;
+import org.openstreetmap.atlas.utilities.configuration.ConfiguredFilter;
 import org.openstreetmap.atlas.utilities.runtime.system.memory.Memory;
 import org.openstreetmap.atlas.utilities.time.Time;
 import org.slf4j.Logger;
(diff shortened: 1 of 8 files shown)


src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -560,6 +561,45 @@ public String getName()
         };
     }
 
+    protected static PairFunction<Tuple2<String, Atlas>, String, Atlas> subatlas(
+            final ConfiguredFilter filter, final AtlasCutType cutType)
+    {
+        return tuple ->
```
matthieun: Shouldn't this be cast as a Serializable function too so Spark can use it?
```suggestion
        return (Serializable & PairFunction<Tuple2<String, Atlas>, String, Atlas>) tuple ->
```

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -560,6 +561,45 @@ public String getName()
...
+                {
+                    subAtlas = subAtlasOptional.get();
+                }
+                else
+                {
+                    subAtlas = null;
+                    logger.error("Unable to extract valid subAtlas code for {}", shardName);
+                }
+            }
+            catch (final Throwable e) // NOSONAR
```
matthieun: In what case would this be a throwable and not an Exception?

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -560,6 +561,45 @@ public String getName()
...
+                // Slice the Atlas
+                final Optional<Atlas> subAtlasOptional = originalAtlas.subAtlas(filter::test,
+                        cutType);
+                if (subAtlasOptional.isPresent())
+                {
+                    subAtlas = subAtlasOptional.get();
+                }
+                else
+                {
+                    subAtlas = null;
```
matthieun: I would make this a `PairFlatMapFunction` so as to return nothing (an empty list) when this is null, and a one item list when this is not. Then there is no risk to have null values in the RDD.
-- Unless I am missing something and null elements are already handled downstream?

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
jwpgage: Spark was able to use the function without the Serializable cast, but I can add it for safety!

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
jwpgage: Good point, I'll change to Exception

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
jwpgage: Null elements are filtered on line 345 of AtlasGenerator!

PR #115, opened 2019-09-04
Atlas loding enhancement

random0909:
### Description:

Added helper functions for Atlas RDD loading. 

The two major ones are: 

`generateCountryShardedAtlasRDD`
`generateExpandedCountryShardedAtlasRDD`

Spark has not been compiled against JDK 11, so the artifact building process will fail.

The solution is temporarily mark the tests as ignored. A separated issue will be created.
Here is the screenshot of the local run to prove that the tests are passed, since the local env can use Java 1.8 with Spark 2.4 

Screen Shot 2019-09-04 at 9 50 51 AM

### Potential Impact:

No impact for downstream since the functions are newly added

### Unit Test Approach:

Used `JavaSparkContext` to simulate the Spark job in the unit test. 

### Test Results:

------


matthieun:
This is stale

PR #122, opened 2019-10-08
AtlasShardVerifier now uses getName instead of toString

lucaspcram:
### Description:
Now that the `Shard#getName` method has been standardized, any code attempting to write a parsable shard string should be using this method.

See: https://github.com/osmlab/atlas/pull/541

### Potential Impact:
Parsing code expecting `toString` to give parsable output may be broken. This fixes that issue.

### Unit Test Approach:
N/A

### Test Results:
N/A

------


lucaspcram:
@matthieun See https://github.com/osmlab/atlas/pull/541#issuecomment-539677506

PR #123, opened 2019-10-14
RDD keys now use JSON instead of ad-hoc strings

lucaspcram:
### Description:
**This pull request depends on https://github.com/osmlab/atlas/pull/546**

Previously, we would key our RDDs using ad-hoc strings like `USA_1-2-3_zz/` or `USA-0_1-2-3`, etc. We now utilize JSON like:
```
{
    "country": "USA",
    "shard": "1-2-3",
    "metadata": {
        "scheme": "zz/"
    }
}
```

### Potential Impact:
Users of the `AbstractMultipleAtlasBasedOutputFormat` will need to adjust their input to respect the new JSON requirements. They may use `PersistenceJsonParser` to make this easier.

### Unit Test Approach:
Full integration tests found in `AtlasGeneratorIntegrationTest`

### Test Results:
All tests pass.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
@@ -353,7 +353,7 @@ public final class AtlasGeneratorHelper implements Serializable
             Memory.printCurrentMemory();
 
             // Output the Name/Atlas couple
-            final String persistenceKey = PersistenceJsonParser.getJsonKey(countryName,
+            final String persistenceKey = PersistenceJsonParser.createJsonKey(countryName,
                     shard.getName(), atlasScheme.getScheme());
             return new Tuple2<>(persistenceKey, atlas);
         };
--- a/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
@@ -17,6 +17,7 @@ import org.apache.spark.broadcast.Broadcast;
 import org.openstreetmap.atlas.exception.CoreException;
 import org.openstreetmap.atlas.generator.persistence.scheme.SlippyTilePersistenceScheme;
 import org.openstreetmap.atlas.generator.tools.caching.HadoopAtlasFileCache;
+import org.openstreetmap.atlas.generator.tools.json.PersistenceJsonParser;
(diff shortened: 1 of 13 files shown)


src/main/java/org/openstreetmap/atlas/generator/tools/json/PersistenceJsonParser.java
```diff
@@ -0,0 +1,169 @@
...
+ *    "shard": "1-2-3",
+ *    "metadata": {
+ *         "scheme": "zz/",
+ *    }
+ * }
+ * </pre>
+ * 
+ * @author lcram
+ */
+public final class PersistenceJsonParser
```
Huyuntj: Seems this parse is trying to read schema file and covert it to json file? if that is the case, check out ObjectMapper, it maybe the thing you need.

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -351,7 +353,9 @@ public String getName()
             Memory.printCurrentMemory();
 
             // Output the Name/Atlas couple
-            return new Tuple2<>(name + Shard.SHARD_DATA_SEPARATOR + atlasScheme.getScheme(), atlas);
+            final String persistenceKey = PersistenceJsonParser.createJsonKey(countryName,
+                    shard.getName(), atlasScheme.getScheme());
+            return new Tuple2<>(persistenceKey, atlas);
```
lucaspcram: @Huyuntj See how the class is being used here.

src/main/java/org/openstreetmap/atlas/generator/tools/json/PersistenceJsonParser.java
lucaspcram: This class exists just to enforce the JSON structure required by `AbstractMultipleAtlasBasedOutputFormat`. So you give it a country, shard, and scheme, and it gives you a JSON key that the output format class understands. I tagged you in a spot where we are using it in atlas-generator. Downstream code should emulate this if they are using saveAsHadoopFile with atlases.

lucaspcram:
@Huyuntj So there is no metadata file. Rather, we are just changing the format of the string key required by the `AbstractMultipleAtlasBasedOutputFormat`. We were running into flexibility issues because we used to stuff tons of information into the and then we would split it on `_` characters. That proved to be too brittle. So now the key is JSON, and the output format parses that JSON. You will see absolutely no change in the names/structure of your atlas output.

PR #124, opened 2019-10-15
Allow for http-based configuration resources

matthieun:
### Description:

Let AtlasGenerator pull its configuration from http-based resources.

### Potential Impact:

This is still backwards compatible, in which it still accepts file-system based resources.

### Unit Test Approach:

TBD

### Test Results:

TBD

------


src/main/java/org/openstreetmap/atlas/generator/sharding/AtlasMissingShardVerifier.java
```diff
@@ -288,7 +283,7 @@ protected int onRun(final CommandMap command)
         {
             return -1;
         }
-        logger.info("Verification ran in: " + fullStart.elapsedSince());
+        logger.info("Verification ran in: {}", fullStart.elapsedSince());
```
adahn6: good catch!

matthieun: All the credit goes to sonar!

PR #128, opened 2019-12-04
Reworked atlas fetcher function to improve performance

adahn6:
### Description:
During relation slicing, the Atlas being sliced is dynamically expanded upon to complete any multipolygon relations. However, until now, the Atlas fetcher function used for this dynamic expansion has been inefficient-- it defaulted to fetching the Atlas for the initial shard from a HadoopAtlasFileCache. This approach ignores that the RDD value for the relation slicing already has the Atlas data for the initial shard stored, so now the fetcher stores that value and will return it directly instead of attempting to load from the HadoopAtlasFileCache.

### Potential Impact:
Fewer HadoopAtlasFileCache calls, so possibly improved performance, but otherwise no impact.

### Unit Test Approach:
N/A

### Test Results:
N/A

------


adahn6:
@lucaspcram yeah, exactly

PR #132, opened 2020-01-10
Fix Json key issue with AtlasStatistics

matthieun:
### Description:

When the RDD keys were switched to JSON in #123 country AtlasStatistics were not upgraded by mistake. The integration test did not fail either since the merging of the shard statistics was taking the last shard statistics object available, making it look valid.

This PR fixes that omission and adds steps to the integration test to make sure it does not happen again.

### Potential Impact:

Country Statistics are back to normal numbers.

### Unit Test Approach:

Updated integration test

### Test Results:

pass

------


matthieun:
Sonarcloud still not catching integration tests. Will reopen when fixed..

PR #142, opened 2020-03-17
Make AtlasStatistics optional

matthieun:
### Description:

The AtlasStatistics portion of AtlasGenerator might not be necessary in all jobs. Here we make it optional with a switch to enable it.

### Potential Impact:

Test jobs will be faster

### Unit Test Approach:

Updated integration test

### Test Results:

Pass

------


src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorParameters.java
```diff
@@ -100,6 +100,9 @@
     public static final Switch<String> CONFIGURED_FILTER_NAME = new Switch<>("configuredFilterName",
             "Name of the filter to be used for configured output", StringConverter.IDENTITY,
             Optionality.OPTIONAL);
+    public static final Switch<Boolean> STATISTICS = new Switch<>("statistics",
+            "Whether to run the shard statistics and country statistics", Boolean::parseBoolean,
+            Optionality.OPTIONAL, "false");
```
adahn6: false is a good default!

PR #147, opened 2020-04-24
Save the lastKey and value first to properly populate exception message

matthieun:
### Description:

When a file system `close()` fails in `AbstractFileOutputFormat`, and the object written is the first one, the error message would print null. This fixes it by setting the temporary objects first before writing them to the file system.

### Potential Impact:

Better logging

### Unit Test Approach:

Code is already covered by tests

### Test Results:

Pass

------


src/main/java/org/openstreetmap/atlas/generator/persistence/AbstractFileOutputFormat.java
```diff
@@ -74,6 +74,9 @@ public void close(final Reporter reporter)
             @Override
             public void write(final String key, final T value)
             {
+
```
lucaspcram, line 77: nit: whitespace

PR #153, opened 2020-05-27
AtlasShardVerifier to work with any FileSystem

matthieun:
### Description:

The AtlasShardVerifier is a command that makes sure that a folder containing Atlas files has all the expected shards. New features are:
- The reference can now be:
  - A file containing a list of expected shards
  - A path to a directory containing Atlas files
- The directory to check does not have to be on the local file system, it can be on any hadoop `FileSystem`

### Potential Impact:

More use cases supported

### Unit Test Approach:

- Added one unit test
- Fixed integration test

### Test Results:

Pass

------

### DIFF
--- a/src/integrationTest/java/org/openstreetmap/atlas/generator/AtlasGeneratorIntegrationTest.java
+++ b/src/integrationTest/java/org/openstreetmap/atlas/generator/AtlasGeneratorIntegrationTest.java
@@ -3,9 +3,12 @@ package org.openstreetmap.atlas.generator;
 import java.io.IOException;
 
 import org.apache.hadoop.fs.Path;
+import org.junit.After;
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 import org.openstreetmap.atlas.exception.CoreException;
+import org.openstreetmap.atlas.generator.tools.spark.SparkJob;
 import org.openstreetmap.atlas.generator.tools.spark.persistence.PersistenceTools;
 import org.openstreetmap.atlas.generator.tools.spark.utilities.SparkFileHelper;
 import org.openstreetmap.atlas.generator.tools.streaming.ResourceFileSystem;
@@ -51,7 +54,14 @@ public class AtlasGeneratorIntegrationTest
     public static final String SLICING_CONFIGURATION = CONFIGURATION
             + "/atlas-relation-slicing.json";
 
-    static
+    @After
+    public void clean()
+    {
+        ResourceFileSystem.clear();
+    }
+
+    @Before
+    public void setup()
     {
         ResourceFileSystem.registerResourceExtractionClass(AtlasGeneratorIntegrationTest.class);
         ResourceFileSystem.addResource(PBF_233, "DMA_cutout.osm.pbf");
@@ -80,6 +90,10 @@ public class AtlasGeneratorIntegrationTest
     @Test
     public void testAtlasGeneration()
     {
+        final StringList sparkConfiguration = new StringList();
+        ResourceFileSystem.simpleconfiguration().entrySet()
+                .forEach(entry -> sparkConfiguration.add(entry.getKey() + "=" + entry.getValue()));
+
         final StringList arguments = new StringList();
         arguments.add("-master=local");
         arguments.add("-output=" + OUTPUT);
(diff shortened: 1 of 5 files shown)


src/integrationTest/java/org/openstreetmap/atlas/generator/AtlasGeneratorIntegrationTest.java
```diff
@@ -51,7 +54,14 @@
     public static final String SLICING_CONFIGURATION = CONFIGURATION
             + "/atlas-relation-slicing.json";
 
-    static
+    @After
+    public void clean()
+    {
+        ResourceFileSystem.clear();
```
adahn6, line 60: hmm what does this do?

matthieun: The ResourceFileSystem is a static hash map of string path to resource objects. It implements the minimum possible of the FileSystem API using the backing of this hash map.  This method clears the static map, so as to look like the file system is empty.
That clean here is also the reason I moved the filling of the file system in the `@Before` method so that it is ready to use before each test

matthieun: The reason why I have done that is I had some random test errors locally, which looked like they came from the fact that the FileSystem was still containing the output of one test when the other test started, which prompted it to fail.

src/main/java/org/openstreetmap/atlas/generator/sharding/AtlasShardVerifier.java
```diff
@@ -56,6 +99,24 @@ protected int onRun(final CommandMap command)
...
+
+    private Set<CountryShard> shardsFromFolder(final String expectedShardsPath,
+            final Map<String, String> sparkConfiguration, final int depth, final PathFilter filter)
+    {
+        logger.trace("listResourcesRecursively: {}", expectedShardsPath);
+        final Time start = Time.now();
+        final Set<CountryShard> result = FileSystemHelper
+                .streamPathsRecursively(expectedShardsPath, sparkConfiguration, filter, depth)
+                .map(Path::getName)
+                .map(name -> StringList.split(name, "/").last()
```
jwpgage, line 114: this name should just be the atlas file name, correct? Not sure why we need to split on "/"

src/test/java/org/openstreetmap/atlas/generator/sharding/AtlasShardVerifierTest.java
```diff
@@ -0,0 +1,91 @@
...
+        ResourceFileSystem.addResource(CHECK + "/ABC/ABC_10-11-14.atlas", atlas);
+        ResourceFileSystem.addResource(CHECK + "/ABC/ABC_10-11-15.atlas", atlas);
+
+        ResourceFileSystem.addResource(INPUT + "/ABC/ABC_10-11-12.atlas", atlas);
+        ResourceFileSystem.addResource(INPUT + "/ABC/ABC_10-11-13.atlas", atlas);
+        ResourceFileSystem.addResource(INPUT + "/ABC/ABC_10-11-14.atlas", atlas);
+
+        new AtlasShardVerifier().runWithoutQuitting(getArguments());
+
+        Assert.assertEquals("ABC_10-11-15",
```
jwpgage, line 69: nice test!

src/main/java/org/openstreetmap/atlas/generator/sharding/AtlasShardVerifier.java
```diff
@@ -1,29 +1,49 @@
...
-            "Folder containing Atlas Shards named by the AtlasGenerator", File::new);
-    private static final Switch<Set<CountryShard>> EXPECTED_SHARDS = new Switch<>("expectedShards",
-            "Text file containing all the expected shards", value -> new File(value).linesList()
-                    .stream().map(CountryShard::forName).collect(Collectors.toSet()));
-    private static final Switch<File> OUTPUT = new Switch<>("output",
-            "The file to list all the missing shards", File::new);
+    public static final Switch<String> ATLAS_FOLDER = new Switch<>("atlasFolder",
+            "Folder containing Atlas Shards named by the AtlasGenerator", StringConverter.IDENTITY,
+            Optionality.REQUIRED);
+    public static final Switch<String> EXPECTED_SHARDS = new Switch<>("expectedShards",
```
jwpgage, line 32: great flexibility here

src/main/java/org/openstreetmap/atlas/generator/sharding/AtlasShardVerifier.java
matthieun: So the map for the atlas file name is two lines below. This one is in case the reference comes from a folder, then the full paths will be included, and we want to strip them out here, if any! If the list comes from a reference file containing a list of shards to expect, then this map here will just do nothing

jwpgage: Hadoop's Path.getName() returns just the filename, not the full path! So you can remove this extra map. https://hadoop.apache.org/docs/current/api/org/apache/hadoop/fs/Path.html

matthieun: Good catch! I will remove it

PR #161, opened 2020-08-27
AtlasRDD to Dataframe conversion methods

jwpgage:
### Description:

A suite of helper methods that convert an atlasRDD into Dataframes, with one Dataframe for each entity type.

### Potential Impact:

Allows for querying atlases within a spark session.

### Unit Test Approach:

~~Test to be added.~~

Test deserialization and data frame creation.

### Test Results:

Dataframes successfully generated.

------


src/main/java/org/openstreetmap/atlas/generator/AtlasDataFrame.java
```diff
@@ -0,0 +1,379 @@
...
+        final int line = 3;
+        final int point = 4;
+        final int relation = 5;
+        switch (value)
+        {
+            case 0:
+                return "NODE";
+            case 1:
+                return "EDGE";
+            case 2:
```
matthieun, line 362: Aah checkstyle! Mabe make it consistent with the ones below?

src/test/java/org/openstreetmap/atlas/generator/AtlasDataFrameTest.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+public class AtlasDataFrameTest
+{
+    private static final Logger logger = LoggerFactory.getLogger(AtlasDataFrameTest.class);
+
+    @Test
+    public void atlasDataFrameTest()
+    {
+        final Resource file = new InputStreamResource(
+                () -> new File("/Users/jamesgage/Desktop/scratch/dfTesting/DMA_9-168-233.atlas")
```
matthieun, line 32: This should become a file in the resources folder. To check it in, make sure it is as small as necessary!

src/main/java/org/openstreetmap/atlas/generator/AtlasDataFrame.java
```diff
@@ -0,0 +1,379 @@
...
+import scala.collection.JavaConversions;
+
+/**
+ * Methods for converting an {@link Atlas} RDD to {@link Dataset}
+ * 
+ * @author jamesgage
+ */
+public final class AtlasDataFrame
+{
+    protected static Dataset<Row> atlasAreasToDataFrame(final JavaRDD<Atlas> atlasRDD,
```
matthieun, line 28: How about making all those methods public?

matthieun (changes requested):
Nice! Just a few nits here and there!
Also it would be nice to have a readme with it that details what this does, as well as defines the schema that each row has!

src/main/java/org/openstreetmap/atlas/generator/AtlasDataFrame.java
jwpgage: yes, definitely a checkstyle-ism :)

lucaspcram (dismissed):

src/test/java/org/openstreetmap/atlas/generator/dataframe/AtlasDataFrameTest.java
```diff
@@ -0,0 +1,69 @@
...
+                javaSparkContext);
+        final Dataset<Row> relationDataframe = AtlasDataFrame
+                .atlasRelationsToDataFrame(atlasJavaRDD, javaSparkContext);
+
+        Assert.assertEquals("24433389000000", areaDataframe.first().get(0));
+        Assert.assertEquals("27989500000000", edgeDataframe.first().get(0));
+        Assert.assertEquals("307351652000000", nodeDataframe.first().get(0));
+        Assert.assertEquals("99202295000000", lineDataframe.first().get(0));
+        Assert.assertEquals("4553243887000000", pointDataframe.first().get(0));
+        Assert.assertEquals("9451753000000", relationDataframe.first().get(0));
```
matthieun, line 66: Is the order of the results in the dataset always deterministic?
Maybe also do a check on the size of each Dataset?

src/main/java/org/openstreetmap/atlas/generator/dataframe/README.md
```diff
@@ -0,0 +1,29 @@
+# `AtlasDataFrame`
+## Overview
+`AtlasDataFrame` is a class that houses methods that convert an atlas RDD to a `DataFrame`. This allows for easily querying the data of an atlas from within an active spark session. Each entity is separated into its own dataframe, and supports its own schema.
+
+### Schemas
+
+Areas:
+
+`identifier` `geometry` `bounds` `tags` `relations`
```
matthieun, line 9: Here I would make that a markdown table, with the type below, like:
`identifier` | `geometry` | `bounds` | `tags` | `relations`
--|--|--|--|--
String|String|String|Map|Array

And the same for the ones below.

src/main/java/org/openstreetmap/atlas/generator/dataframe/README.md
```diff
@@ -0,0 +1,29 @@
+# `AtlasDataFrame`
```
matthieun, line 1: Can you add a relative link to this page here? https://github.com/osmlab/atlas/blob/dev/README.md#whats-in-it

src/test/java/org/openstreetmap/atlas/generator/dataframe/AtlasDataFrameTest.java
matthieun: Might also be nice to test the inside of the whole row... on top of the ID.

src/main/java/org/openstreetmap/atlas/generator/dataframe/README.md
jwpgage: Does the relative link need to be from within the same repo? It seems that all the others are contained within atlas

src/test/java/org/openstreetmap/atlas/generator/dataframe/AtlasDataFrameTest.java
jwpgage: Added a check for size and the tags, but haven't tested the long strings for geometry etc

src/main/java/org/openstreetmap/atlas/generator/dataframe/README.md
matthieun: Ah right! Different repo haha! Maybe we need to have it linked in the atlas-generator's main readme?

matthieun: That's probably for another PR though. We need to add more discoverability there..

matthieun (approved):

lucaspcram (approved):

PR #162, opened 2020-09-03
Replace some non-inclusive language

adahn6:
### Description:
This update replaces some non-inclusive language in the codebase. The master branch is replaced by the main branch in travis and circleci scripts.

### Potential Impact:
One parameter has been renamed from 'master' to 'cluster', and accordingly any use of that parameter should be updated to reflect that. 

### Test Results:

------


src/main/java/org/openstreetmap/atlas/generator/tools/spark/SparkJob.java
```diff
@@ -53,7 +53,7 @@
             StringConverter.IDENTITY, Optionality.OPTIONAL);
     public static final Switch<String> OUTPUT = new Switch<>("output",
             "Output path of the Spark Job", StringConverter.IDENTITY, Optionality.REQUIRED);
-    public static final Switch<String> MASTER = new Switch<>("master", "The spark master URL",
+    public static final Switch<String> CLUSTER = new Switch<>("cluster", "The spark cluster URL",
```
matthieun, line 56: I think we should keep both, and have logic below to find which one was populated... That way downstream scripts do not fail right away?

config/checkstyle/checkstyle.xml
```diff
@@ -143,7 +143,7 @@
 		<module name="TrailingComment">
 			<property name="legalComment" value="NOSONAR"/>
 		</module>
-		<module name="TodoComment" />
+		<!--<module name="TodoComment" />-->
```
matthieun, line 146: I think we'd want to keep that to be consistent with Atlas, and instead use `// TODO` in place of `// @TODO` which triggers it.

matthieun (changes requested):
Thanks! Just two small requests!

config/checkstyle/checkstyle.xml
adahn6: oh, okay! I can fix that

src/main/java/org/openstreetmap/atlas/generator/tools/spark/SparkJob.java
adahn6: hmm.... I don't think that's a great idea, because there's no way of marking a parameter deprecated. so from a user standpoint, if they're not reading our code, it'll seem just as sudden when we finally remove the parameter

adahn6: I guess we could update the description string, but it seems pretty unlikely any downstream scripts are going to read that

matthieun: Ok, we can leave this as is then!

matthieun (approved):

PR #167, opened 2020-11-03
Spark FileSystems should be closed when no longer needed.

tsmock:
This fixes a resource leak, which will improve resource usage.

### Description:

Put `org.apache.hadoop.fs.FileSystem` creation into `try-with-resource` blocks. This ensures that the `FileSystem` is properly closed. There are also some fixes for sonar sprinkled in (mostly deduplication of strings).

### Potential Impact:

Major impact: Less resource utilization
No impact: No new exceptions are thrown (those that are thrown through the try-with-resources are caught and handled silently, to avoid breaking downstream users).

### Unit Test Approach:

N/A

### Test Results:

With https://github.com/osmlab/atlas-checks/pull/398, I was able to reduce the memory required to run the test check on NZL from 15 GB to 1GB. It may have been possible to run it with less memory (so < 1GB).

Notes:
* Additional modifications to `atlas-checks` was required in order to run the current `dev` branch of `atlas-generator`. Specifically, the resolution strategy for `jackson` had to be commented out (due to 54d1bd2b607f430379e18e2fee9166f6d803bf94 ).

------

In doubt: [Contributing Guidelines](https://github.com/osmlab/atlas/blob/dev/CONTRIBUTING.md)

TODO list:
List of everything that uses `org.apache.hadoop.fs.FileSystem` (from imports):

* ~src/main/java/org/openstreetmap/atlas/generator/PbfLocator.java~ [DONE, implements AutoCloseable]
* ~src/main/java/org/openstreetmap/atlas/generator/persistence/AbstractFileOutputFormat.java~ [N/A -- FileSystem passed into class]
* ~src/main/java/org/openstreetmap/atlas/generator/persistence/MultipleAtlasCountryStatisticsOutputFormat.java~ [N/A -- FileSystem passed into class]
* ~src/main/java/org/openstreetmap/atlas/generator/persistence/MultipleAtlasFeatureChangeOutput.java~ [N/A -- FileSystem passed into class]
* ~src/main/java/org/openstreetmap/atlas/generator/persistence/MultipleAtlasGeoJsonOutputFormat.java~ [N/A -- FileSystem passed into class]
* ~src/main/java/org/openstreetmap/atlas/generator/persistence/MultipleAtlasOutputFormat.java~ [N/A -- FileSystem passed into class]
* ~src/main/java/org/openstreetmap/atlas/generator/persistence/MultipleAtlasProtoOutputFormat.java~ [N/A -- FileSystem passed into class]
[...]

PR #171, opened 2021-01-19
AutoCloseable: Use autocloseable resources appropriately

tsmock:
### Description:

Add Closeable Resources, for use when a Resource depends upon an AutoCloseable resource that must not be closed prior to the user finishing with the resource.

### Potential Impact:

This largely comes down to memory impact. In osmlab/atlas-generator#167, closing resources significantly freed up space and sped up processing. Unfortunately, during testing, a bug where resources were closed prior to reads was missed. This fixes that issue by creating resources that store the closeable resources internally, which can then be closed automatically when finished.

This may have been missed due to specifying a `-sharding` parameter and/or a less congested network.

### Test Results:

I've done several runs with an executor on a Raspberry Pi (also running a hadoop server) with two worker nodes on Mac computers.

This depends upon ~https://github.com/osmlab/atlas/pull/727~ (just merged, build won't pass until version bump)

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/generator/tools/spark/SparkJob.java
+++ b/src/main/java/org/openstreetmap/atlas/generator/tools/spark/SparkJob.java
@@ -29,10 +29,9 @@ import org.openstreetmap.atlas.generator.tools.spark.converters.SparkOptionsStri
 import org.openstreetmap.atlas.generator.tools.spark.persistence.PersistenceTools;
 import org.openstreetmap.atlas.generator.tools.spark.utilities.SparkFileHelper;
 import org.openstreetmap.atlas.streaming.compression.Decompressor;
-import org.openstreetmap.atlas.streaming.resource.AbstractResource;
 import org.openstreetmap.atlas.streaming.resource.FileSuffix;
-import org.openstreetmap.atlas.streaming.resource.InputStreamResource;
-import org.openstreetmap.atlas.streaming.resource.Resource;
+import org.openstreetmap.atlas.streaming.resource.InputStreamResourceCloseable;
+import org.openstreetmap.atlas.streaming.resource.ResourceCloseable;
 import org.openstreetmap.atlas.streaming.resource.http.GetResource;
 import org.openstreetmap.atlas.utilities.conversion.StringConverter;
 import org.openstreetmap.atlas.utilities.runtime.Command;
@@ -79,19 +78,31 @@ public abstract class SparkJob extends Command implements Serializable
 
     private transient JavaSparkContext context;
 
-    public static Resource resource(final String path, final Map<String, String> configurationMap)
+    /**
+     * Get resources from a path
+     *
+     * @param path
+     *            The resource path
+     * @param configurationMap
+     *            The configuration for {@link FileSystemCreator#get(String, Configuration)}
+     * @return A {@link ResourceCloseable} (this should be closed on finish)
+     */
+    public static ResourceCloseable resource(final String path,
+            final Map<String, String> configurationMap)
     {
         if (path.startsWith("http"))
         {
             return new GetResource(path);
         }
-        try (FileSystem fileSystem = new FileSystemCreator().get(path, configurationMap))
+        // This is closed by the ResourceCloseable class
+        final FileSystem fileSystem = new FileSystemCreator().get(path, configurationMap);
+        try
         {
             if (!fileSystem.exists(new Path(path)))
             {
                 return null;
             }
-            final AbstractResource resource = new InputStreamResource(() ->
+            final InputStreamResourceCloseable resource = new InputStreamResourceCloseable(() ->
             {
                 try
                 {
@@ -101,7 +112,7 @@ public abstract class SparkJob extends Command implements Serializable
                 {
                     throw new CoreException("Unable to open {}", path, e);
                 }
-            });
+            }, fileSystem);
             if (path.endsWith(FileSuffix.GZIP.toString()))
             {
                 resource.setDecompressor(Decompressor.GZIP);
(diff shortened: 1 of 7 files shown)


tsmock:
I need to bump the Atlas version for this, as osmlab/atlas#727 just got merged, but it is ready for review (despite the error).

src/main/java/org/openstreetmap/atlas/generator/tools/spark/SparkJob.java
```diff
@@ -79,19 +78,31 @@
...
+    public static ResourceCloseable resource(final String path,
+            final Map<String, String> configurationMap)
     {
         if (path.startsWith("http"))
         {
             return new GetResource(path);
         }
-        try (FileSystem fileSystem = new FileSystemCreator().get(path, configurationMap))
+        // This is closed by the ResourceCloseable class
+        final FileSystem fileSystem = new FileSystemCreator().get(path, configurationMap);
```
matthieun, line 98: With that change, we would never close FileSystems anymore, just resources?

src/main/java/org/openstreetmap/atlas/generator/tools/spark/persistence/PersistenceTools.java
```diff
@@ -43,8 +46,24 @@ public PersistenceTools(final Map<String, String> configurationMap)
...
-                .resource(SparkFileHelper.combine(input, BOUNDARIES_FILE), this.configurationMap));
+        try (ResourceCloseable resource = SparkJob
+                .resource(SparkFileHelper.combine(input, BOUNDARIES_FILE), this.configurationMap))
+        {
+            return CountryBoundaryMap.fromPlainText(resource);
+        }
+        catch (final Exception e)
+        {
+            logger.error(e);
+            if (e instanceof RuntimeException)
```
matthieun, line 57: I would remove this check, since CoreException is also a runtime exception. I would just wrap everything in the core exception below

src/main/java/org/openstreetmap/atlas/generator/tools/spark/sharded/ShardedSparkJob.java
```diff
@@ -88,8 +90,15 @@ public void start(final CommandMap command)
...
-        final CountryBoundaryMap worldBoundaries = new CountryBoundaryMapArchiver()
-                .read(resource(countryShapes));
+        final CountryBoundaryMap worldBoundaries;
+        try (ResourceCloseable resource = resource(countryShapes))
+        {
+            worldBoundaries = new CountryBoundaryMapArchiver().read(resource);
+        }
+        catch (final Exception e)
+        {
+            throw new CoreException(e.getMessage());
```
matthieun, line 100:
```suggestion
            throw new CoreException("Could not read {}", countryShapes, e);
```

src/main/java/org/openstreetmap/atlas/generator/tools/spark/persistence/PersistenceTools.java
```diff
@@ -43,8 +46,24 @@ public PersistenceTools(final Map<String, String> configurationMap)
...
+            {
+                throw (RuntimeException) e;
+            }
+            else
+            {
+                throw new CoreException("Could not close {}", e,
+                        SparkFileHelper.combine(input, BOUNDARIES_FILE));
```
matthieun, line 64:
```suggestion
                throw new CoreException("Could not close {}",
                        SparkFileHelper.combine(input, BOUNDARIES_FILE), e);
```

matthieun (changes requested):
Thanks for the addition! Just a few suggestions related to exception handling!

src/main/java/org/openstreetmap/atlas/generator/tools/spark/SparkJob.java
tsmock: This is a fair point. There are a few paths where the path doesn't exist, or there is an exception thrown, where the filesystem isn't closed. I'll fix this.

src/main/java/org/openstreetmap/atlas/generator/tools/spark/persistence/PersistenceTools.java
tsmock: I wanted to avoid wrapping a RuntimeException in another exception (i.e., I wanted to avoid cases where there is a CoreException with 14 inner CoreExceptions).

src/main/java/org/openstreetmap/atlas/generator/tools/spark/persistence/PersistenceTools.java
tsmock: I see how this works. Is it a good idea to do that though? The `CoreException` constructor for that is undocumented, so the behavior that uses the _last_ argument as the cause could change at any time. I'd rather use the explicit constructor where I know that the `Throwable` is in the appropriate location. If this change is truly desired, then maybe docs should indicate that the `CoreException(String, Object...)` is preferred over `CoreException(String, Throwable, Object...)`. I'll open up a PR for that if needed.

src/main/java/org/openstreetmap/atlas/generator/tools/spark/persistence/PersistenceTools.java
tsmock: Just to clarify, did you still want me to change this, or can I leave this as is?

src/main/java/org/openstreetmap/atlas/generator/tools/spark/persistence/PersistenceTools.java
tsmock: Ping.

PR #176, opened 2021-02-09
Mutator: Allow specific fetchers to expand across country codes

matthieun:
### Description:

By default, a mutator spark job will use dynamicAtlas policies which expand to neighboring shards within the same country, according to the provided country boundaries.

This update allows a user to toggle `{"countryvore": "true"}` when configuring an Atlas fetcher, which will prompt it to find shards across country boundaries. If a Shard crosses multiple countries, the fetcher will make sure to multi-atlas all those files together (same shard, but different countries) in one single Atlas which contains it all.

### Potential Impact:

This is a new feature, so minimal impact. The potential is for slightly reduced performance with very large sharding schemes, since it requires broadcasting a map of all shards to countries they overlap to all the executors.

### Unit Test Approach:

Added a few unit tests, and some integration test.

### Test Results:

So far all integration tests pass, including the new one that expands across countries. To be sure that the expansion is symmetric, it computes the area of a multipolygon that it can only get right if proper shard expansion happened.

------


lucaspcram:
What use case would benefit from this feature?

adahn6:
this looks great! 🥳

matthieun:
@lucaspcram The main use case is to allow porting country slicing to the mutator platform, as a mutation. It currently sits within the AtlasGenerator spark job.
Being able to multi-atlas shards from many countries is a pre-requisite to be able to run country slicing on large relations, the members of which might not all sit in the same countries.

adahn6 (approved):

src/integrationTest/java/org/openstreetmap/atlas/mutator/AbstractAtlasMutatorIntegrationTest.java
```diff
@@ -52,9 +52,12 @@
     public static final String OUTPUT_BOUNDARIES_META = OUTPUT + "/" + BOUNDARY_META_NAME;
     public static final String OUTPUT_COUNTRY_LIST = OUTPUT + "/" + AtlasMutator.COUNTRY_AND_LEVELS;
 
-    public static void addResource(final String path, final String name)
+    public static void addAtlasResource(final String path, final Atlas atlas)
     {
-        addResource(path, name, false);
+        final long size = 8192;
```
lucaspcram, line 57: I like to do
```java
final long size = 1024 * 8;
```
Makes it clearer that this number is supposed to be 8 KiB. But this is a minor point.

src/integrationTest/java/org/openstreetmap/atlas/mutator/AtlasMutatorCountryvoreIntegrationTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.openstreetmap.atlas.generator.AtlasGeneratorParameters;
+import org.openstreetmap.atlas.geography.atlas.Atlas;
+
+/**
+ * @author matthieun
+ */
+public class AtlasMutatorCountryvoreIntegrationTest extends AbstractAtlasMutatorIntegrationTest
```
lucaspcram, line 16: Pointing out `Countryvore` which I assume is a typo. If not, then I would add a comment explaining what it means!

lucaspcram (commented):
Looks good! A couple soft comments that I don't think are blocking merging.

src/integrationTest/java/org/openstreetmap/atlas/mutator/AbstractAtlasMutatorIntegrationTest.java
matthieun: This was already there, must have been shuffled around, but I will fix that.

src/integrationTest/java/org/openstreetmap/atlas/mutator/AtlasMutatorCountryvoreIntegrationTest.java
matthieun: Haha! ok!

lucaspcram (approved):

PR #185, opened 2021-05-27
Remove countries with no boundaries.

atiannicelli:
### Description:

This switches the default boundary file used in aws script and also remove countries that do not have any boundaries in the boundary file.

### Potential Impact:

Will allow Java11 AWS generation to complete. Without this the generation fails on countries without boundaries.

### Unit Test Approach:

Executed manually and all regions worked


atiannicelli:
@jwpgage 
If you have time would you mind reviewing this patch and if you are good with it merging it? I think it is pretty quick and easy.

PR #186, opened 2021-07-08
Added an OSM Node filter to AtlasMissingShardVerifier

lucaspcram:
### Description:
Added a by-ID filter for OSM Nodes to the `AtlasMissingShardVerifier`. This will allow the verifier to ignore problematic nodes that builders of atlases may not want to ingest, and therefore reduce false positives in the verifier's output.

There are some arrangement diffs here too, it's been a long time since anyone has checked on this class 😂

### Potential Impact:
Potential for fewer false positives in verifier output.

### Unit Test Approach:
Tested by running locally

### Test Results:
👍

------


jwpgage (approved):
Excellent addition!

matthieun (approved):

PR #187, opened 2021-07-21
Add keepall parameter to aws script

atiannicelli:
### Description:
In preparation for the keepall flag to be added to the repo, this simply adds a parameter to the AWS script so we can indicate that we want keepall or not. 

### Potential Impact:
None unless you use it. If you specify the flag on a jar that doesn't support the -keepAll flag then it will fail.
This requires #186 to be merged.

### Unit Test Approach:
I ran a script run with and without the parameter specified and the proper parameter was passed to the EMR cluster.

### Test Results:
All good.
------


atiannicelli:
@tsmock FYI

tsmock (approved):
LGTM.

jwpgage (approved):

PR #188, opened 2021-08-06
Enable relation geometries

adahn6:
### Description:

This PR adds support for Relation geometries added in https://github.com/osmlab/atlas/pull/764 . The major new addition here is a step after raw Atlas generation in which geometric Relation members are consolidated into RDD pairs for their relation id, built, then re-applied to the raw Atlases and saved. This step, known as "augmented Atlases", allows for geometric Relations to have all members across all PBFs built assigned to them quickly and efficiently, so that every shard can have a deterministic, complete geometry for the Relation *without* having to load in every relevant Atlas.

### Potential Impact:

There is a lot that changes with this! But, a significant amount of work has been done to preserve most forward-facing APIs such that there are minimally necessary changes to users. For the most part, the default Atlas generation job will now work faster and with less memory, *and* generate stored geometries for when any relevant Atlas API is called on them.

### Unit Test Approach:

Limited, to be honest most relevant cases are covered in Atlas

### Test Results:

------


src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -172,6 +186,92 @@ public String getName()
...
+    }
+
+    /**
+     * This function takes a given tuple of shard and atlas to the set of multipolygons that were
+     * assembled for it. It will then apply each of those relation multipolygons as feature changes
+     * and save the final atlas with the geometries serialized
+     *
+     * @return function that does the above
+     */
+    protected static PairFunction<Tuple2<String, Tuple2<Atlas, org.apache.spark.api.java.Optional<Set<Tuple2<Long, MultiPolygon>>>>>, String, Atlas> augmentAtlas()
```
matthieun, line 212: I am quite surprised by the optional here. Shouldn't the set just be empty?

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -172,6 +186,92 @@ public String getName()
...
+    {
+        return (Serializable & PairFunction<Tuple2<String, Tuple2<Atlas, org.apache.spark.api.java.Optional<Set<Tuple2<Long, MultiPolygon>>>>>, String, Atlas>) tuple ->
+        {
+            final Atlas rawAtlas = tuple._2._1;
```
matthieun, line 216: `_2` is a direct access to the member. We usually try to use the accessor methods instead in Java (`_2()`), like below!

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -172,6 +186,92 @@ public String getName()
...
+            if (!tuple._2()._2.isPresent())
```
matthieun, line 217: With just the empty set, here you could just check that the set is empty or not...

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
```diff
@@ -427,28 +599,71 @@ public String getName()
             }
             catch (final Throwable e) // NOSONAR
             {
-                throw new CoreException(ERROR_MESSAGE,
-                        AtlasGeneratorJobGroup.WAY_SECTIONED_PBF.getDescription(), countryShardName,
-                        e);
+                logger.warn("Sectioning had error, probably was null atlas",
```
matthieun, line 602: Error, then logger.error? Or that happens too often?

src/main/java/org/openstreetmap/atlas/mutator/filtering/ChangeFilter.java
```diff
@@ -532,16 +547,23 @@ private void populateRelationShapes(final Change originalChange, final long rela
...
+                .filter(featureChange -> geometricRelationChanges(featureChange, change))
+                // Collect
+                .collect(Collectors.toList());
         final List<FeatureChange> filteredChanges = change.changes()
                 // Strip shallow feature changes
                 .filter(featureChange -> featureChange.afterViewIsFull() || this.atlas
                         .entity(featureChange.getIdentifier(), featureChange.getItemType()) != null)
                 // Strip second degree edges
-                .filter(overlap(initialShardBounds, change))
+                .filter(featureChange -> overlap(initialShardBounds, change).test(featureChange)
```
matthieun, line 560: Might be nicer here to filter twice for readability, since it is an &&

src/main/java/org/openstreetmap/atlas/mutator/filtering/ChangeFilter.java
```diff
@@ -557,10 +579,14 @@ private void populateRelationShapes(final Change originalChange, final long rela
                 .stream(new MultiIterable<>(this.atlas.edges(), this.atlas.relations()))
                 .filter(entity -> !filteredChangeIdentifiers
                         .contains(AtlasEntityKey.from(entity.getType(), entity.getIdentifier())))
-                .filter(entity -> !overlap(initialShardBounds, change)
-                        // Here this first FeatureChange is meaningless, just used to be able to
-                        // call the overlaps method on the specific atlas entity.
-                        .test(FeatureChange.add(CompleteEntity.from(entity))))
+                .filter(entity -> !filteredGeometricRelationChanges.stream()
+                        .anyMatch(featureChange -> featureChange.getItemType().equals(ItemType.EDGE)
+                                && featureChange.getIdentifier() == entity.getIdentifier())
```
matthieun, line 584: I would do the same here...

matthieun (changes requested):
Thanks! Just a few nits!
- Very good commenting! Makes it much easier to follow such a complex change
- The changeFilter updates look good at first glance, but I am not sure 100%. We'd need to see it work at scale first to confirm!

atiannicelli (commented):
Don't you need to add

    maven { url "https://repo.eclipse.org/content/repositories/jts-snapshots" }

to build.gradle to be able to build?

adahn6:
> Don't you need to add
> 
> ```
> maven { url "https://repo.eclipse.org/content/repositories/jts-snapshots" }
> ```
> 
> to build.gradle to be able to build?
> 
> I guess if you build atlas first from atlas repo and then depend on the atlas version then it works without this line. But if you want to be able to dynamically pick up atlas repo changes in "../atlas" then you might want to add that line above to the build.gradle in atlas-generator.
> 
> If you don't want to go that way I get it.

this is a good question. ideally, we'll keep the inclusion of the snapshot portion of jts short lived-- it should release by the end of the month-- so I think we should avoid adding any further direct inclusion of it here.

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
adahn6: it's an input from leftOuterJoin() :/

src/main/java/org/openstreetmap/atlas/generator/AtlasGeneratorHelper.java
adahn6: good catch!

lucaspcram (approved):

PR #193, opened 2021-11-04
Fix bug in PbfLoader

adahn6:
### Description:

PbfLoader accidentally regenerates raw atlases instead of using the sub atlas optional 💀 

### Potential Impact:

Data quality should be better limited to shard bounds, atlas generation time will drop since we don't do it twice anymore

### Unit Test Approach:

No unit test since this actually behaved as expected with respect to data generation, it's just doing it twice unnecessarily.

### Test Results:

------


matthieun (approved):
🥳

PR #195, opened 2021-12-01
pbfShardGenerator: Add bounds to extracted pbf files

tsmock:
### Description:

This is mostly so that software can see the bounds without having to
figure out what shard the file (i.e., so we can more easily look at an
area in JOSM and figure out what is going on around the boundary).


atiannicelli (approved):
Awesome

PR #196, opened 2021-12-16
PersistenceTools: log4j -> slf4j logging

tsmock:
Move to slf4j logging throughout atlas generator.

I should have used slf4j initially, anyway.

There should be no impact on downstream projects.

See also
* https://github.com/osmlab/atlas-checks/pull/644
* https://github.com/osmlab/atlas/pull/780

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,8 +6,8 @@ plugins {
     id 'signing'
     id 'checkstyle'
     id 'jacoco'
-    id "com.diffplug.gradle.spotless" version "3.27.0"
-    id 'org.sonarqube' version '2.8'
+    id "com.diffplug.spotless" version "6.1.0"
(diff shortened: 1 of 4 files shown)


tsmock:
Note: We can use dependency substitution to use the osgeo -norce version instead of excluding osgeo.

atiannicelli:
log4j 2.16.0 creates another vulnerability (CVE-2021-45105). So we need to upgrade to 2.17.0

tsmock:
> log4j 2.16.0 creates another vulnerability ([CVE-2021-45105](https://github.com/advisories/GHSA-p6xc-xr62-6r2g)). So we need to upgrade to 2.17.0

`atlas-generator` does not _explicitly_ depend upon Log4J. It is visible from `atlas`.

adahn6 (approved):

PR #204, opened 2022-06-07
pbf-reshard script

vladlemberg:
### Description:

create script for additional sharding if sharding.txt is outdated.
1) search pbf release folder for shards that larger than 5mb(default).
2) create osmium json config file.
3) execute sharding
4) update release folder, update sharding.txt

### Potential Impact:

sharding strategy will be up to date with the latest data size. 

### Unit Test Approach:

none

### DIFF
--- a/scripts/cloudPBFShardControl/pbfReShard.py
+++ b/scripts/cloudPBFShardControl/pbfReShard.py
@@ -3,6 +3,8 @@
 
 """
 this script will search pbf release folder for shards that above certain size (default 5mb) and re-shards them locally.
+@author: Vladimir Lemberg
+@history: 06/10/2022 Created
 """
 import argparse
 import json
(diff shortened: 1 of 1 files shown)


scripts/cloudPBFShardControl/pbfReShard.py
```diff
@@ -0,0 +1,315 @@
...
+        create new sharding.txt file
+        """
+        if self.change:
+            logger.info("creating new sharding tree file")
+            shard_date = date.today().strftime("%Y%m%d")
+            new_sharding_quadtree_file = f"sharding_quadtree_{shard_date}.txt"
+            with open(os.path.join(self.tmpDir, new_sharding_quadtree_file), 'w') as output_file:
+                output_file.writelines(self.shardingContent)
+            # copy original sharding file
+            copy_command(self.shardingFileName, os.path.join(self.releasePbfDir, f"sharding_original_{shard_date}.txt"))
```
atiannicelli, line 228: Three things:

1. Don't you want to use the filename that was passed in here?
2. we need to add the hour and minute to this date string.
3. I would tack the data on the end of the filename.

In other words, if the user passes in:
`special_sharding_file.txt`
then we want the script to copy it to:
`special_sharding_file.txt.202206100327`

atiannicelli (changes requested):

scripts/cloudPBFShardControl/pbfReShard.py
vladlemberg: resolved.

atiannicelli (approved):
Love it!

adahn6 (approved):

